PR #2, opened 2020-06-26
Allow choice for student/teacher signup

evantey14:
- Made enum for signup type
- Made buttons for switching type

Next step would be to insert different forms based on signup type.


client/src/App.tsx
```diff
@@ -9,7 +9,7 @@ import Nav from "./layout/Nav";
 import Footer from "./layout/Footer";
 
 import LoginPage from "./login/LoginPage";
-import SignupPage from "./login/SignupPage";
+import SignupPage from "./signup/SignupPage";
```
wishdasher, line 12: I want these to be in the same package for later. You could rename it to something better than `login` if you want.

wishdasher:
For frontend commits with a requested review, it would be nice to include a screenshot in here (in the PR discussion section) to show what the new component looks like.

client/src/signup/SignupPage.tsx
```diff
@@ -0,0 +1,69 @@
+import React, { Component } from "react";
+import SignupForm from "./SignupForm";
+import { UserState } from "../App";
+import { SignupType } from "./types";
+
+type SignupProps = {
```
wishdasher, line 6: You could just rename these to `Props` and `State`, since it's pretty clear what they're meant to be used for. If we need to export the type, then we can give it a more specific name.

client/src/signup/types.ts
```diff
@@ -0,0 +1,6 @@
+enum SignupType {
+  Student,
+  Teacher,
+}
+
+export { SignupType };
```
wishdasher, line 6: I love types files!

client/src/signup/SignupPage.tsx
```diff
@@ -0,0 +1,69 @@
...
+  onLogin: (data: UserState) => void;
+  loggedIn: boolean;
+  username: string;
+};
+
+type SignupState = {
+  signupType: SignupType;
```
wishdasher, line 13: Perhaps a more descriptive name? `chosenSignupType` or something similar?

client/src/signup/SignupPage.tsx
```diff
@@ -0,0 +1,69 @@
...
+class SignupPage extends Component<SignupProps, SignupState> {
+  constructor(props: SignupProps) {
+    super(props);
+    this.state = {
+      signupType: SignupType.Student,
+    };
+  }
+
+  getButtonClass(signupType: SignupType) {
+    return this.state.signupType === signupType ? "button is-active" : "button";
```
wishdasher, line 25: Personal preference:
```
`button ${this.state.signupType === signupType ? "is-active" : ""}`
```

client/src/signup/SignupPage.tsx
wishdasher: I like `selected` better, actually

PR #4, opened 2020-06-28
Refactor student reg

mvadari:
Merging in all of the comments from @evantey14


client/src/registration/StudentRegDashboard.tsx
```diff
@@ -14,9 +15,31 @@ type State = {
   liabilityCheck: boolean;
   medliabCheck: boolean;
   updateProfileCheck: boolean;
-  regStatus: string;
+  regStatus: RegStatusOption;
 };
 
+// helper functions
```
evantey14, line 21: So we don't have to fix this here, but I think I agree with a past comment from Sophie that I prefer verb-named functions (renderTextInSection, renderTaskWithLink, etc).

server/core/models/__init__.py
```diff
@@ -89,18 +132,25 @@ class Meta:
...
+    teacherreg = models.ForeignKey(TeacherRegistration, on_delete=models.CASCADE)
     clazz = models.ForeignKey(Class, on_delete=models.CASCADE)
 
-    # ensures a teacher can't register the same class twice
     class Meta:
-        unique_together = (("teacher", "clazz"),)
+        unique_together = (("teacherreg", "clazz"),)
+
+    @property
+    def student(self):
```
evantey14, line 148: I think this should be teacher? (same in the __str__)

PR #5, opened 2020-06-29
Studentreg forms

mvadari:
Update Profile form
* Populated with backend information
* Submits information back to the backend, checks it off as completed on submit
image

Dummy forms for all other forms
* Checks it off as completed in the backend on submit
image

All assorted API calls to process data


client/package.json
```diff
@@ -7,6 +7,7 @@
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.3.2",
     "@testing-library/user-event": "^7.1.2",
+    "@types/reach__router": "^1.3.5",
```
wishdasher, line 10: Could you add this as a dev dependency instead? You'll have to remove this one and do an `add --dev`.

client/src/App.tsx
```diff
@@ -15,6 +15,9 @@ import Home from "./info/Home";
 import Dashboard from "./dashboard/Dashboard";
 
 import RegDashboard from "./registration/RegDashboard";
+import UpdateProfileForm from "./registration/UpdateProfileForm";
+import EmergencyInfoForm from "./registration/EmergencyInfoForm";
+import DummyForm from "./registration/DummyForm";
```
wishdasher, line 20: Nit: Please alphabetize. Maybe we can add a lint for that? @evantey14

client/src/App.tsx
```diff
@@ -109,13 +112,41 @@ class App extends Component<{}, State> {
...
+              isStudent={this.state.isStudent}
+              url="waiver"
+              formName="Liability Waiver Form"
+            />
+            <DummyForm
+              path="/:program/:edition/availability"
+              isStudent={this.state.isStudent}
+              url="availability"
+              formName="Program Availability"
+            />
```
wishdasher, line 149: All of these parts of registration should be under the same route, probably `/program/:edition/reg`. We can show the correct form depending on which step of the registration process they are at. Okay to add a TODO for now.

Also perhaps worth looking into how we can structure the routes. Certainly, not all of them will be in `App.tsx` ultimately.

client/src/App.tsx
```diff
@@ -109,13 +112,41 @@ class App extends Component<{}, State> {
...
               isTeacher={this.state.isTeacher}
             />
+            <UpdateProfileForm
+              path="/:program/:edition/updateprofile"
+              isStudent={this.state.isStudent}
+              isTeacher={this.state.isTeacher}
+            />
+            <EmergencyInfoForm
+              path="/:program/:edition/emergencyinfo"
+              isStudent={this.state.isStudent}
```
wishdasher, line 130: Why does this have an `isStudent` prop?

client/src/registration/DummyForm.tsx
```diff
@@ -0,0 +1,64 @@
...
+import axiosInstance from "../axiosAPI";
+import { navigate } from "@reach/router";
+
+type Props = {
+  isStudent: boolean;
+  url: string;
+  formName: string;
+  program: string;
+  edition: string;
+};
```
wishdasher, line 11: Nit: alphabetize.

client/src/registration/DummyForm.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+type State = {};
+
+// This is a placeholder for forms that haven't been implemented yet
+// All they do is mark the form as completed in the backend on submit
+
+class DummyForm extends React.Component<Props, State> {
```
wishdasher, line 18: Nit: Alternatively, do an `import React, { Component}` and only have `extends Component` here. I don't really care which one we do (we've been doing the former), but I don't want a mix.

client/src/registration/DummyForm.tsx
```diff
@@ -0,0 +1,64 @@
...
+  };
+
+  render() {
+    return (
+      <div className="container">
+        <div className="columns">
+          <div className="column is-6 is-offset-3">
+            <h1 className="has-text-centered is-size-3">{this.props.formName}</h1>
+            <form
+              onSubmit={(e: React.FormEvent<HTMLFormElement>) => this.handleSubmit(e, this.state)}
```
wishdasher, line 43: You should be able to do `onSubmit={this.handleSubmit}` and pull `data` in that method from `this.state`? (Not sure why you're passing state at all.)

client/src/registration/DummyForm.tsx
```diff
@@ -0,0 +1,64 @@
...
+            >
+              <label className="label" htmlFor="name">
```
wishdasher, line 45: Remove `htmlFor` here; it's not doing anything.

client/src/registration/EmergencyInfoForm.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+type State = {
+  something: string;
+};
+
+function isValidField(prop: string, obj: State): prop is keyof State {
+  return prop in obj;
+}
+
+class EmergencyInfoForm extends React.Component<Props, State> {
```
wishdasher, line 19: Same re import.

client/src/registration/EmergencyInfoForm.tsx
```diff
@@ -0,0 +1,133 @@
...
+  isStudent: boolean;
+  program: string;
+  edition: string;
+};
+
+type State = {
+  something: string;
+};
+
+function isValidField(prop: string, obj: State): prop is keyof State {
```
wishdasher, line 15: I don't know this return type syntax. Is it not just `boolean`?

client/src/registration/EmergencyInfoForm.tsx
```diff
@@ -0,0 +1,133 @@
...
+        },
+      })
+      .then(res => {
+        this.setState({});
+      });
+  }
+
+  handleChange = (e: React.FormEvent<HTMLInputElement>) => {
+    const name = e.currentTarget.name as string;
+    const value = e.currentTarget.value as string;
```
wishdasher, line 46: Do you need the `as string`? I don't remember if event types had typed fields...

client/src/registration/UpdateProfileForm.tsx
```diff
@@ -0,0 +1,326 @@
+import React from "react";
+import axiosInstance from "../axiosAPI";
+import { navigate } from "@reach/router";
+
+type Props = {
+  isStudent: boolean;
+  isTeacher: boolean;
+  program: string;
+  edition: string;
```
wishdasher, line 9: Nit: alphabetize.

client/src/registration/UpdateProfileForm.tsx
```diff
@@ -0,0 +1,326 @@
...
+  lastName: string;
+  phoneNumber: string;
+  pronouns: string;
+  city: string;
+  state: string;
+  country: string;
+  dateOfBirth: string;
+  gradYear: string;
+  school: string;
+  affiliation: string;
```
wishdasher, line 24: Same here.

client/src/registration/UpdateProfileForm.tsx
```diff
@@ -0,0 +1,326 @@
...
+};
+
+const PronounOptions: string[] = [
+  "He/Him/His",
+  "She/Her/Hers",
+  "They/Them/Theirs",
+  "Other",
+  "Prefer Not to Say",
```
wishdasher, line 32: Extract these out to some file, it could be in the same directory for now. Add a
`TODO: consider placing these in a profile package`.

client/src/registration/UpdateProfileForm.tsx
```diff
@@ -0,0 +1,326 @@
...
+];
+
+function isValidField(prop: string, obj: State): prop is keyof State {
+  return prop in obj;
```
wishdasher, line 36: Same comments from other file.

client/src/registration/UpdateProfileForm.tsx
```diff
@@ -0,0 +1,326 @@
...
+          {icon && (
+            <span className="icon is-left">
+              <i className={"fas fa-" + icon}></i>
+            </span>
+          )}
+          {help && <p className="help"> {help} </p>}
+        </div>
+      </div>
+    );
+  }
```
wishdasher, line 210: I would pull all these out into a separate file, and into a separate `forms` package.

client/src/registration/UpdateProfileForm.tsx
```diff
@@ -0,0 +1,326 @@
...
+          {label}
+        </label>
+        <div className={className}>
+          <div className="select is-normal is-expanded">
+            <select
+              id={name}
+              name={name}
+              value={value}
+              placeholder={label}
+              onChange={this.handleChange}
```
wishdasher, line 194: Can we do `onBlur` instead of `onChange`? This is probably up for debate, though. Non-validated inputs, like name, should be updated `onBlur`. For valididated inputs, it can probably vary, depending on how frequently we want to trigger an API call. @evantey14 

Okay to put a TODO here for now.

client/src/registration/UpdateProfileForm.tsx
wishdasher: Also I appreciate these helper functions, but the number of parameters is an issue...will continue to think about it.

client/src/registration/UpdateProfileForm.tsx
```diff
@@ -0,0 +1,326 @@
...
+              {this.props.isStudent &&
+                this.labeledInput(
+                  "Current School",
+                  "school",
+                  this.state.school,
+                  "text",
+                  "chalkboard-teacher"
+                )}
+
+              {this.props.isTeacher &&
```
wishdasher, line 302: Yeah, I would rather have separate update profile forms than have `props.isStudent` and `props.isTeacher`, especially when we expand to more user types. Thoughts, @evantey14?

server/core/views/api.py
```diff
@@ -101,3 +98,199 @@ def current_studentreg(request):
     studentreg, _ = StudentRegistration.objects.get_or_create(student=user, program=prog)
     # TODO figure out how to change the default regstatus based on the program's status
     return Response(StudentRegSerializer(studentreg).data)
+
+
+class Profile(APIView):
```
wishdasher, line 103: @evantey14 to review this change.

package.json
```diff
@@ -3,13 +3,18 @@
...
   "scripts": {
     "postinstall": "cd client && yarn install --production=false",
     "build": "cd client && yarn build"
   },
   "engines": {
     "node": "14.x",
     "yarn": "1.x"
+  },
+  "dependencies": {
+    "@types/reach__router": "^1.3.5"
```
wishdasher, line 18: This shouldn't be here at all. Only the one in `client/package.json` should be.

client/package.json
mvadari: Done

client/src/App.tsx
mvadari: Done (well not the linter but)

client/src/App.tsx
mvadari: I'm not really sure I understand how they can all have the same URL. How do we differentiate between them? Maybe the dashboard requires a bit of a discussion, because I was thinking that you click on the links in order to access the form, and the link you click on shows you that form. 

Added a TODO about routes

client/src/App.tsx
mvadari: I was thinking some sort of redirect/error message if you're not a student (since you shouldn't be able to access the page)

client/src/registration/DummyForm.tsx
mvadari: I did it, but I feel like it makes more sense when things are grouped on how they're used.

client/src/registration/DummyForm.tsx
mvadari: Done - that was an accident (seems to have not been caught in the other forms, fixed that)

client/src/registration/DummyForm.tsx
mvadari: Done - also wasn't caught in the other forms, has been fixed

client/src/registration/DummyForm.tsx
mvadari: Done

client/src/registration/EmergencyInfoForm.tsx
mvadari: It appears to be how you declare that `prop` is a key in type `obj`. I'm not really sure how it works, but some Googling shows that this is the commonly touted solution.

client/src/registration/EmergencyInfoForm.tsx
mvadari: This was remnants of me debugging random failures. Removed.

client/src/registration/UpdateProfileForm.tsx
mvadari: Done

client/src/registration/UpdateProfileForm.tsx
mvadari: Done

client/src/registration/UpdateProfileForm.tsx
mvadari: Created a `forms` folder, put it in `constants.ts`

client/src/registration/UpdateProfileForm.tsx
mvadari: Same comments from before

client/src/registration/UpdateProfileForm.tsx
mvadari: Done

I guess for the booleans you could create alternate functions for True vs False (e.g. disabledCustomInput or something)

client/src/registration/UpdateProfileForm.tsx
mvadari: Put a TODO and added onBlur just in case

client/src/registration/UpdateProfileForm.tsx
mvadari: I feel like for more user types we might want the same form and have the same conditional elements

package.json
mvadari: Done

client/src/App.tsx
evantey14: I'll add a trello card

client/src/registration/UpdateProfileForm.tsx
evantey14: lol yeah idk the answer to this probably not too many calls?

client/src/registration/UpdateProfileForm.tsx
evantey14: I think separate forms is probably more readable but I also don't think it's a big deal for now since we don't have too many fields.

client/src/App.tsx
wishdasher: The same route would lead to a page that would load each of those components. I made this a trello card for now, so I'm marking resolved.

client/src/App.tsx
wishdasher: Per the previous comment, this route wouldn't exist if we have a single registration route.

client/src/registration/EmergencyInfoForm.tsx
wishdasher: I'm still not sure why this is a separate function.

client/src/registration/UpdateProfileForm.tsx
wishdasher: I don't think those constants should live in `forms`. If anything, let's put them into a global constants file.

client/src/registration/UpdateProfileForm.tsx
```diff
@@ -0,0 +1,326 @@
...
+        program: this.props.program,
+        edition: this.props.edition,
+      })
+      .then(result => {
+        /*TODO check validity of submitted data*/
+        navigate("dashboard");
+      });
+  };
+
+  customInput(
```
evantey14: Do we want to make this method a verb rather than a noun?

client/src/registration/UpdateProfileForm.tsx
```diff
@@ -0,0 +1,326 @@
...
+        {icon && (
+          <span className="icon is-small is-left">
+            <i className={"fas fa-" + icon}></i>
+          </span>
+        )}
+      </div>
+    );
+  }
+
+  labeledInput(
```
evantey14: same comment as above

server/core/urls.py
```diff
@@ -61,11 +61,20 @@
...
+    path("api/emergency_info/", views.EmergencyInfo.as_view()),
+    path("api/medliab/", views.MedicalLiability.as_view()),
+    path("api/waiver/", views.LiabilityWaiver.as_view()),
+    path("api/availability/", views.Availability.as_view()),
+    # auth API calls
     path("api/add_user/", views.CreateUser.as_view()),
+    path("api/current_user/", views.current_user),
     path("api/token/", jwt_views.TokenObtainPairView.as_view(), name="token_obtain_pair"),
     path("api/token/refresh/", jwt_views.TokenRefreshView.as_view(), name="token_refresh"),
+    # for heroku
```
evantey14, line 78: ah yes, thanks

server/core/views/api.py
```diff
@@ -101,3 +98,199 @@ def current_studentreg(request):
...
+        }
+        if user.is_student:
+            profile = user.student_profile
+            data["date_of_birth"] = profile.date_of_birth
+            data["grad_year"] = profile.grad_year
+            data["school"] = profile.school
+
+        if user.is_teacher:
+            profile = user.teacher_profile
+            data["affiliation"] = profile.affiliation
```
evantey14, line 132: So first, `user.profile` should return the correct profile (so the student_profile and teacher_proiles calls aren't necessary).

Second, I think you can do this in python with something like
```
data = {
    "first_name": user.first_name,
     # other user fields
     **profile
}
```
or if that doesn't work, try looking into dict's update function?

server/core/views/api.py
```diff
@@ -101,3 +98,199 @@ def current_studentreg(request):
...
+
+        return Response(data)
+
+    def post(self, request, format=None):
```
evantey14, line 136: Let's add some validation here or at least drop a TODO for what we want to validate.

server/core/views/api.py
```diff
@@ -101,3 +98,199 @@ def current_studentreg(request):
...
+        user = request.user
+        data = request.data
+
+        user.first_name = data["first_name"]
+        user.last_name = data["last_name"]
+        user.email = data["email"]
```
evantey14, line 142: I think in python we prefer data.get("email") over data["email"] because the former returns None (or whatever default we tell it to) and the latter throws a KeyError

client/src/registration/UpdateProfileForm.tsx
```diff
@@ -0,0 +1,326 @@
...
+  return prop in obj;
+}
+
+class UpdateProfileForm extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      email: "",
+      firstName: "",
+      lastName: "",
```
evantey14: what exactly do we want students to be able to update in this form?

I think I originally was thinking it should be just profile info (not email/name) -- this would also simplify backend calls since we only have to deal with profile objects.

server/core/views/api.py
```diff
@@ -101,3 +98,199 @@ def current_studentreg(request):
...
+        profile.pronouns = data["pronouns"]
+        profile.city = data["city"]
+        profile.state = data["state"]
+        profile.country = data["country"]
+        profile.school = data["school"]
+        profile.save()
+
+        if "update_profile" in data and data["update_profile"]:
+            assert "program" in data
+            assert "edition" in data
```
evantey14, line 166: these are where dictionary .get functions are useful -- you can then test that they're not None.

server/core/views/api.py
```diff
@@ -101,3 +98,199 @@ def current_studentreg(request):
...
+        if user.is_student:
+            self.update_student(user, data)
+
+        if user.is_teacher:
+            self.update_teacher(user, data)
+
+        # TODO add correction checks
+        return Response({"message": "Success!"})
+
+    def update_student(self, user, data):
```
evantey14, line 154: I think this function seems to be doing 2 things? The first half updates the profile, the second marks it as done. We should add a comment/TODO that these should be done atomically. also it might be worth abstracting them into smalller functions for readability ("update_student" doesn't really tell me what happens in this function).

client/src/App.tsx
mvadari: Ah I see

client/src/App.tsx
mvadari: Ah ok. I'll leave it in for now since this is going to be refactored later anyways.

server/core/views/api.py
```diff
@@ -101,3 +98,199 @@ def current_studentreg(request):
...
+        profile.affiliation = data["affiliation"]
+        profile.save()
+
+        if "update_profile" in data and data["update_profile"]:
+            assert "program" in data
+            assert "edition" in data
+            program = Program.objects.get(name=data["program"], edition=data["edition"])
+            teacherreg = TeacherRegistration.objects.get(teacher=user, program=program)
+            teacherreg.update_profile_check = True
+            teacherreg.save()
```
evantey14, line 188: Is there a good reason to do this stuff now? I think we want to keep PRs fairly small -- in which case it'd be better to have this just focus on student profile updates (the PR is technically named Studentreg forms). 

We can think about how to incorporate teachers later -- it might be better to have 2 separate classes that inherit some shared abstract class? Atm this code kinda feels like two parallel classes just jammed into one class?

server/core/views/api.py
```diff
@@ -101,3 +98,199 @@ def current_studentreg(request):
...
+
+
+class EmergencyInfo(APIView):
+    """
```
evantey14, line 192: maybe change this comment so it describes the current state of this (something about how it's currently a shell).

server/core/views/api.py
```diff
@@ -101,3 +98,199 @@ def current_studentreg(request):
...
+
+        studentreg.emergency_info_check = True
+        studentreg.save()
+
+        # TODO add correction checks
+        return Response({"message": "Success!"})
+
+
+class MedicalLiability(APIView):
+    """
```
evantey14, line 228: Same deal -- improve comment

client/src/registration/EmergencyInfoForm.tsx
mvadari: I think the `prop is keyof State` also does some part of the analysis. 

I tried refactoring so the function wasn't necessary, and I couldn't figure it out. If you can, then we can delete it. But when I was first writing out the form code, it wouldn't work without it.

server/core/views/api.py
```diff
@@ -101,3 +98,199 @@ def current_studentreg(request):
...
+
+        studentreg.medliab_check = True
+        studentreg.save()
+
+        # TODO add correction checks
+        return Response({"message": "Success!"})
+
+
+class LiabilityWaiver(APIView):
+    """
```
evantey14, line 252: ^^^

server/core/views/api.py
```diff
@@ -101,3 +98,199 @@ def current_studentreg(request):
...
+
+        studentreg.liability_check = True
+        studentreg.save()
+
+        # TODO add correction checks
+        return Response({"message": "Success!"})
+
+
+class Availability(APIView):
+    """
```
evantey14, line 276: ^^^

client/src/registration/EmergencyInfoForm.tsx
mvadari: https://dev.to/kingdaro/indexing-objects-in-typescript-1cgi

client/src/registration/UpdateProfileForm.tsx
mvadari: Done

client/src/registration/UpdateProfileForm.tsx
mvadari: (y) made this just student specific then

server/core/views/api.py
```diff
@@ -101,3 +98,199 @@ def current_studentreg(request):
     studentreg, _ = StudentRegistration.objects.get_or_create(student=user, program=prog)
     # TODO figure out how to change the default regstatus based on the program's status
     return Response(StudentRegSerializer(studentreg).data)
+
+
+class Profile(APIView):
+    """
+    Create a new user.
```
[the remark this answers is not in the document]
mvadari: Made the comment more descriptive

client/src/registration/UpdateProfileForm.tsx
mvadari: Done

client/src/registration/UpdateProfileForm.tsx
mvadari: Done

server/core/views/api.py
mvadari: I was doing this for now since root is a teacher and a student, and I was lazy. But yeah that's probably better, and we should start creating student and teacher roots. 

It looks like `**profile` only works if the object is mappable, but I can't figure out how to make that possible.

mvadari: nvm figured it out

server/core/views/api.py
mvadari: Added TODOs

server/core/views/api.py
mvadari: Wouldn't we want it to throw a KeyError? There shouldn't be any field that somehow isn't there.

client/src/registration/UpdateProfileForm.tsx
mvadari: tbh I haven't looked at the current version on the website. I assumed it would be good to allow them to edit that easily, given the number of people who have schedules under their parents' names every program.

server/core/views/api.py
mvadari: Isn't that the same concept? What's the benefit of one over the other?

server/core/views/api.py
mvadari: Took it out for now. Just did it to be holistic.

server/core/views/api.py
mvadari: Made it atomic (there's a decorator in django to do this for us)

Split into 2 functions that seem reasonable to me

server/core/views/api.py
mvadari: Done oops

server/core/views/api.py
mvadari: Done oops

server/core/views/api.py
mvadari: Done oops

server/core/views/api.py
mvadari: Done oops

client/src/registration/UpdateProfileForm.tsx
evantey14: Hmm alright I guess it's okay to leave for now but I think for future things of similar complexity we want to keep this as simple as possible. something something MVP -- if the feature isn't that important, I'd prefer cleaner code.

Anyhow, I'll mark as resolved (and lol I guess I can steal this for signup stuff)

server/core/views/api.py
evantey14: I think it's smoother to have the gets return None and an if statement that catches it.

the alternative to handling the KeyError would be something like a try-except block, but if we put it in the wrong place / add more code to this function it gets more and more likely that the try-except catches a KeyError that isn't this specific KeyError which could make it hard to debug

server/core/views/api.py
evantey14: Again it's laying the groundwork for good error throwing in the future. `assert`s always throw Assertion errors -- but python default errors are not the most descriptive. Ideally down the line we'd make our own error classes specific to our issues with useful error messages. 

anytime `"field" in dict` is used it's almost immediately followed by `dict["field"]` if the former was True -- get() means you don't have to deal with later accesses and your conditional test is still as simple as `is None`

mvadari:
> I think everything looks good on my end. I'd prefer we do error checking with .get for dicts, but otherwise I think it looks fine.
All right, done

PR #6, opened 2020-06-29
Display class schedule

mvadari:
Add class schedule display for everything after the lottery
image
Add skeleton of a class changes dashboard 
image

### DIFF
--- a/client/src/dashboard/StudentDashboard.tsx
+++ b/client/src/dashboard/StudentDashboard.tsx
@@ -66,7 +66,7 @@ export default class StudentDashboard extends Component<Props, State> {
             {this.state.previousPrograms.map((program, index) => {
               return (
                 <h3 className="is-size-5" key={index}>
-                  {program.name}: <a href={program.url + "/dashboard"}>View</a>
+                  {program.name}: <a href={`${program.url}/dashboard`}>View</a>
                 </h3>
               );
             })}
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -22,6 +22,14 @@ function renderTextInSection(displayedText: string, centered = false) {
   );
 }
 
+function renderLinkedText(displayedText: string, url: string) {
+  return (
+    <h3 className="is-size-5 has-text-centered">
+      <a href={url}>{displayedText}</a>
+    </h3>
+  );
+}
+
 class StudentRegDashboard extends Component<Props, State> {
   constructor(props: Props) {
     super(props);
@@ -106,6 +114,7 @@ class StudentRegDashboard extends Component<Props, State> {
           {this.renderClassStatus()}
           {this.renderClassCatalog()}
         </div>
+        {renderLinkedText("Back to Dashboard", "dashboard")}
       </div>
     );
   }
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -60,34 +60,27 @@ class StudentRegDashboard extends Component<Props, State> {
   }
 
   renderClassSchedule() {
-    // tables might not be very accessible
-    return (
-      <div className="table-container">
-        <table className="table is-fullwidth is-striped is-hoverable">
-          <thead>
-            <th>Time</th>
-            <th>Class</th>
-          </thead>
-          <tbody>
-            {this.state.classes.map((clazz, index) => {
-              return (
-                <tr>
-                  <th>{this.state.timeslots[index]}</th>
-                  <td>{clazz}</td>
-                </tr>
-              );
-            })}
-          </tbody>
-        </table>
-      </div>
-    );
-  }
-
-  renderClassStatus() {
     return (
       <div className="column is-6">
         <h2 className="has-text-centered is-size-3">Classes</h2>
-        {this.renderClassSchedule()}
+        <div className="table-container">
+          <table className="table is-fullwidth is-striped is-hoverable">
+            <thead>
+              <th>Time</th>
+              <th>Class</th>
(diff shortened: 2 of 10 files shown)


client/src/dashboard/StudentDashboard.tsx
```diff
@@ -54,7 +54,9 @@ export default class StudentDashboard extends Component<Props, State> {
               return (
                 <h3 className="is-size-5" key={index}>
                   {program.name}:{" "}
-                  <a href={program.url}>{program.registered ? "Go to Dashboard" : "Register!!"}</a>
+                  <a href={program.url + "/dashboard"}>
```
wishdasher, line 57: Alternatively `${program.url}/dashboard`. I personally find that easier to read.

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -0,0 +1,116 @@
...
+  timeslots: string[];
+  classes: string[];
+};
+
+// helper functions
+
+//mostly used for placeholders
+function renderTextInSection(displayedText: string, centered = false) {
+  return (
+    <h3 className={"is-size-5" + (centered ? " has has-text-centered" : "")}>{displayedText}</h3>
```
wishdasher, line 21: What is the `has` class?

client/src/registration/StudentRegDashboard.tsx
```diff
@@ -50,39 +63,56 @@ class StudentRegDashboard extends Component<Props, State> {
...
+        },
+      })
+      .then(res => {
+        this.setState({
+          availabilityCheck: res.data.availability_check,
+          emergencyInfoCheck: res.data.emergency_info_check,
+          liabilityCheck: res.data.liability_check,
+          medliabCheck: res.data.medliab_check,
+          updateProfileCheck: res.data.update_profile_check,
+          regStatus: res.data.reg_status,
```
wishdasher, line 91: I wonder if we should be storing a `data` object in the state rather than each of these individually.

client/src/registration/StudentRegDashboard.tsx
```diff
@@ -50,39 +63,56 @@ class StudentRegDashboard extends Component<Props, State> {
...
+        });
+      });
   }
 
   renderRegStatus() {
     return (
-      <div className="column is-4">
-        <h2 className="has-text-centered is-size-3">Registration Status</h2>
+      <div className="column is-3">
+        <h2 className="has-text-centered is-size-3  mb-2">Registration Status</h2>
```
wishdasher, line 115: One space please.

client/src/registration/StudentRegDashboard.tsx
```diff
@@ -105,20 +135,63 @@ class StudentRegDashboard extends Component<Props, State> {
...
+        {renderTextInSection("View class schedule here")}
+      </div>
+    );
+  }
+
+  renderClassView() {
     switch (this.state.regStatus) {
       case RegStatusOption.ClassPreferences:
-        return renderTextInSection("Change class preferences here");
+        return this.renderClassPrefs(true);
```
wishdasher, line 184: I feel like there's a better way of abstracting whether edits are allowed than having this boolean here, which is really opaque.

server/config/settings.py
```diff
@@ -145,7 +145,11 @@
     "PAGE_SIZE": 10,
 }
 
-CORS_ORIGIN_WHITELIST = ("http://localhost:3000",)
+CORS_ORIGIN_WHITELIST = (
+    "http://localhost:3000",
```
wishdasher, line 149: We don't want to allow `http`?

server/core/views/api.py
```diff
@@ -104,3 +104,34 @@ def current_studentreg(request):
     studentreg, _ = StudentRegistration.objects.get_or_create(student=user, program=prog)
     # TODO figure out how to change the default regstatus based on the program's status
     return Response(StudentRegSerializer(studentreg).data)
+
+
+class StudentClasses(APIView):
+    """
+    Get/update a student's class list.
```
wishdasher, line 111: This needs a more detailed docstring. Does this get all classes ever? Why are we not just retrieving the classes for a specific program?

client/src/dashboard/StudentDashboard.tsx
mvadari: Done

client/src/registration/ChangeClassesDashboard.tsx
mvadari: Oops typo

client/src/registration/StudentRegDashboard.tsx
mvadari: What's the tradeoff?

client/src/registration/StudentRegDashboard.tsx
mvadari: Oops typo - I assumed the linter would catch things like this?

client/src/registration/StudentRegDashboard.tsx
mvadari: I did a thing that I think helps with this

server/config/settings.py
mvadari: This was me attempting to get the heroku frontend to connect to the heroku backend, rather than my local backend. But removed.

server/core/views/api.py
mvadari: Oops docstring needed to be more specific, it's per-program

server/config/settings.py
mvadari: Update: localhost doesn't understand how to do HTTPS, I think it might be a docker thing? paging @evantey14

evantey14: oof yeah i need to figure out the heroku config stuff

PR #7, opened 2020-07-01
Add script and sample data for populating database

wishdasher:
With alias, use `dockex python setup_sample_data.py` to run.

Feel free to modify the sample data or add more so we have a useful set to use when testing features and elements.


server/setup_sample_data.py
```diff
@@ -0,0 +1,67 @@
...
+    StudentRegistration,
+    TeacherClassRegistration,
+    TeacherProfile,
+    TeacherRegistration,
+)
+
+
+DATA_DIR = "sample_data"
+models_user = [ESPUser, StudentProfile, TeacherProfile]
+models_program = [
```
evantey14, line 27: is there a reason for these names as opposed to `user_models` and `program_models`?

server/setup_sample_data.py
```diff
@@ -0,0 +1,67 @@
...
+]
+
+
+# If fields are references to other model's primary keys, process that here
+def process_data(model, data):
+    print(data)
+    for field, value in data.items():
+        ref_model = model._meta.get_field(field).related_model
+        if ref_model is not None:
+            ref_obj = globals()[ref_model.__name__].objects.get(pk=value)
```
evantey14, line 43: whats the whole `globals` thing doing?

server/setup_sample_data.py
```diff
@@ -0,0 +1,67 @@
...
+    Class,
+    StudentRegistration,
+    StudentClassRegistration,
+    TeacherRegistration,
+    TeacherClassRegistration,
+]
+
+
+# If fields are references to other model's primary keys, process that here
+def process_data(model, data):
```
evantey14, line 38: maybe let's put a doc string here to explain what this function does more specifically. it might be good to include what args you expect and what kind of thing is being returned.

what kind of processing is done? is it just doing fk lookups?

server/setup_sample_data.py
wishdasher: Too used to going general > specific I guess.

server/setup_sample_data.py
wishdasher: Nothing, good catch

server/setup_sample_data.py
wishdasher: Made things more specific. Withheld a docstring as this is not, and is not meant to be, a public method.

PR #9, opened 2020-07-01
Api reorganization

mvadari:
Move all API URLs in the frontend to a separate file, so you don't have to go hunting for the URLs to edit them
Change the API URL hierarchy to follow standard REST API practice ([source](https://medium.com/hashmapinc/rest-good-practices-for-api-design-881439796dc9)) ([hierarchy](https://docs.google.com/document/d/1lgPza4EP7eM_kbxA9F6Uxt7tgZgLjZsbVQwhNVYJUS4/edit))
Rename/create more permissions (some of these aren't really used because I can't figure out how)

### DIFF
--- a/server/core/urls.py
+++ b/server/core/urls.py
@@ -6,7 +6,7 @@ from rest_framework_simplejwt import views as jwt_views
 from . import views
 
 router = routers.DefaultRouter()
-router.register("teacherprograms", views.TeacherProgramViewSet)
+router.register("dashboard/teacher/", views.TeacherProgramViewSet)
 
 app_name = "core"
 urlpatterns = [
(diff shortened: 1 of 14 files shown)


server/core/views/api.py
```diff
@@ -207,6 +198,14 @@ def post(self, request, format=None):
         # TODO add correction checks
         return Response({"message": "Success!"})
 
+    def get_object(self, program=None, edition=None):
```
evantey14, line 201: why are `program` and `edition` defaulted to `None`?

also is there a particular reason this is named `get_object`? if not, I think it might be nice to have a more specific name that better describes what we're getting

server/core/views/api.py
```diff
@@ -215,23 +214,29 @@ class MedicalLiability(APIView):
...
 
         # TODO add correction checks
         return Response({"message": "Success!"})
 
+    def get_object(self, program=None, edition=None):
+        user = self.request.user
+        program = Program.objects.get(name=program, edition=edition)
+        # TODO add check for if studentreg object exists
+        studentreg = StudentRegistration.objects.get(student=user, program=program)
+        self.check_object_permissions(self.request, studentreg)
```
evantey14, line 236: why is this line necessary? (particularly if it wasn't necessary in emergency info?

server/core/views/api.py
```diff
@@ -215,23 +214,29 @@ class MedicalLiability(APIView):
...
         # TODO maybe add other permissions to this so you can't accidentally do this?
+        # Like something from formstack
 
         studentreg.medliab_check = True
         studentreg.save()
 
         # TODO add correction checks
         return Response({"message": "Success!"})
 
+    def get_object(self, program=None, edition=None):
```
evantey14, line 231: same comments as in emergency info

server/core/views/api.py
```diff
@@ -257,6 +258,15 @@ def post(self, request, format=None):
...
```
evantey14, line 261: same comment as above --- honestly, there's probably a better place to put this code so it's not written 3 times

server/core/views/api.py
```diff
@@ -265,23 +275,27 @@ class Availability(APIView):
...
+    def post(self, request, program, edition, format=None):
+        studentreg = self.get_object(program, edition)
 
         studentreg.availability_check = True
         studentreg.save()
 
         # TODO add correction checks
         return Response({"message": "Success!"})
 
+    def get_object(self, program=None, edition=None):
```
evantey14, line 291: same comments as above

server/core/urls.py
```diff
@@ -6,7 +6,7 @@
 from . import views
 
 router = routers.DefaultRouter()
-router.register("teacherprograms", views.TeacherProgramViewSet)
+router.register("dashboard/teacher/", views.TeacherProgramViewSet)
```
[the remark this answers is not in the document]
mvadari: Viewsets work differently than views. I think viewsets can be made to work the same, but it's a little more complicated to get them to do the thing, so it was easier to just do it separately. It'll be less weird once all the other Django URLs are gone.

server/core/views/api.py
mvadari: It's overriding a method, and [appears to be common practice](https://www.django-rest-framework.org/tutorial/3-class-based-views/)

server/core/views/api.py
mvadari: I don't quite understand how object permissions work, I need to look more into that. I'll take out the line and add a TODO.

PR #11, opened 2020-07-03
Flesh out admin panel a lot so it's more useful for admins

mvadari:
* Add more information in the list display for each model
* Add more information in the edit display for each model
* Add search-by fields to search for models more effectively
* Only show what is necessary for model creation
* Make some parts of models uneditable because it doesn't make sense to edit them (e.g. editing the student or class in a StudentClassRegistration object - you should delete it and create a new one)
* Add list filters (e.g. by program, by status, by user type)
* Add inlines for Student/TeacherClassRegistration objects in Class, StudentReg, TeacherReg
** Make sure the options for these inlines correspond to the correct program
* Reorganize the classes into an admin folder

TODO add actions so that some things can be done en masse (e.g. changing StudentReg status)

Adding some screenshots (but not all)
image
image
image
image
image

### DIFF
--- a/server/core/admin/__init__.py
+++ b/server/core/admin/__init__.py
@@ -8,6 +8,7 @@ from core.models import (
 )
 from django.contrib import admin
 
+from .users import *  # noqa
 from .utils import ActiveProgramFilter
 
 
--- a/server/core/admin/__init__.py
+++ b/server/core/admin/__init__.py
@@ -107,5 +107,17 @@ class TeacherRegistrationAdmin(admin.ModelAdmin):
     list_filter = (ActiveProgramFilter,)
 
 
-admin.site.register(TeacherClassRegistration)
-admin.site.register(StudentClassRegistration)
+@admin.register(TeacherClassRegistration)
+class TeacherClassRegistrationAdmin(admin.ModelAdmin):
+    readonly_fields = ("teacherreg", "teacher", "program", "clazz")
+    list_display = ("teacher", "program", "clazz")
+    search_fields = ("teacherreg", "clazz")
+    fields = ("teacher", "program", "clazz")
+
+
+@admin.register(StudentClassRegistration)
+class StudentClassRegistrationAdmin(admin.ModelAdmin):
+    readonly_fields = ("studentreg", "student", "program", "clazz")
+    list_display = ("student", "program", "clazz")
+    search_fields = ("studentreg", "clazz")
+    fields = ("student", "program", "clazz")
--- a/server/core/admin/__init__.py
+++ b/server/core/admin/__init__.py
@@ -8,14 +8,9 @@ from core.models import (
 )
 from django.contrib import admin
 
+from .filters import ActiveProgramClassRegFilter, ActiveProgramFilter
+from .inlines import StudentClassRegistrationInline, TeacherClassRegistrationInline
 from .users import *  # noqa
-from .utils import (
-    ActiveProgramClassRegFilter,
-    ActiveProgramFilter,
-    ProgramActiveFilter,
-    StudentClassRegistrationInline,
-    TeacherClassRegistrationInline,
-)
 
 
 @admin.register(Class)
@@ -47,7 +42,6 @@ class ClassAdmin(admin.ModelAdmin):
 @admin.register(Program)
 class ProgramAdmin(admin.ModelAdmin):
     list_display = ("__str__", "student_reg_open", "student_reg_status", "teacher_reg_open")
-    list_filter = (ProgramActiveFilter, "student_reg_open", "teacher_reg_open")
 
     def save_model(self, request, obj, form, change):
         if "student_reg_status" in form.changed_data:
--- a/server/core/admin/__init__.py
+++ b/server/core/admin/__init__.py
@@ -1,67 +1,14 @@
-from django.contrib import admin
-
-from .models import (
+from core.models import (
     Class,
-    ESPUser,
     Program,
     StudentClassRegistration,
-    StudentProfile,
     StudentRegistration,
     TeacherClassRegistration,
-    TeacherProfile,
     TeacherRegistration,
 )
+from django.contrib import admin
 
-
-class ActiveProgramFilter(admin.SimpleListFilter):
-    # Human-readable title which will be displayed in the
-    # right admin sidebar just above the filter options.
-    title = "program"
-
-    # Parameter for the filter that will be used in the URL query.
-    parameter_name = "program"
-
-    def lookups(self, request, model_admin):
-        """
-        Returns a list of tuples. The first element in each
-        tuple is the coded value for the option that will
-        appear in the URL query. The second element is the
-        human-readable name for the option that will appear
-        in the right sidebar.
-        """
-        active_programs = Program.get_active_programs()
-        return [(program.id, str(program)) for program in active_programs]
-
-    def queryset(self, request, queryset):
-        """
-        Returns the filtered queryset based on the value
-        provided in the query string and retrievable via
-        `self.value()`.
-        """
-        if self.value():
-            return queryset.filter(program=self.value())
-        else:
-            return queryset
-
-
-class StudentInline(admin.StackedInline):
-    model = StudentProfile
-
-
-class TeacherInline(admin.StackedInline):
-    model = TeacherProfile
-
-
-@admin.register(ESPUser)
-class ESPUserAdmin(admin.ModelAdmin):
-    fields = (
-        "username",
-        ("first_name", "last_name"),
-        "email",
-    )
-    list_display = ("username", "email", "first_name", "last_name", "is_student", "is_teacher")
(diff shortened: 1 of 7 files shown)


server/core/admin/utils.py
```diff
@@ -0,0 +1,135 @@
...
+        active_programs = Program.get_active_programs()
+        return [(program.id, str(program)) for program in active_programs]
+
+    def queryset(self, request, queryset):
+        """
+        Returns the filtered queryset based on the value
+        provided in the query string and retrievable via
+        `self.value()`.
+        """
+        if self.value():
```
evantey14, line 69: I'm not sure I understand what exactly this whole self.value thing is? I think this function is particularly unreadable...not sure if you have any ideas for how to help readers better understand what's going on?

server/core/admin/utils.py
```diff
@@ -0,0 +1,135 @@
...
+
+class ProgramActiveFilter(admin.SimpleListFilter):
+    title = "active"
+    parameter_name = "active"
+
+    def lookups(self, request, model_admin):
+        return [(1, "Yes"), (0, "No")]
+
+    def queryset(self, request, queryset):
+        print("(***************)")
```
evantey14, line 83: nit: remove

server/core/models/__init__.py
```diff
@@ -58,13 +58,20 @@ def get_current_student_programs(user):
         current_programs = Program.objects.filter(id__in=current_studentregs)
         return current_programs
 
+    @staticmethod
```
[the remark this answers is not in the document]
evantey14: Also I guess I'm not sure if this is the definition we want for active programs?

server/core/admin/utils.py
```diff
@@ -0,0 +1,135 @@
+from core.models import (
+    Program,
+    StudentClassRegistration,
+    StudentProfile,
+    TeacherClassRegistration,
+    TeacherProfile,
+)
+from django.contrib import admin
+
+# Filters
```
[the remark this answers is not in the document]
evantey14: honestly I think i would prefer a `filters.py` and `inlines.py`

server/core/admin/utils.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class ActiveProgramFilter(admin.SimpleListFilter):
```
evantey14, line 13: I think it's a bit confusing to have an `ActiveProgramFilter` and a `ProgramActiveFilter`. Also I'm not super convinced that we need the program filters (generally active programs, ones with open student/teacher reg are just gonna be near the top).

server/core/admin/utils.py
```diff
@@ -0,0 +1,135 @@
...
+    def queryset(self, request, queryset):
+        value = self.value()
+        if value == "student":
+            students = StudentProfile.objects.all().values_list("user", flat=True)
+            return queryset.filter(id__in=students)
+        elif value == "teacher":
+            teachers = TeacherProfile.objects.all().values_list("user", flat=True)
+            return queryset.filter(id__in=teachers)
+        elif value == "admin":
+            return queryset.filter(is_superuser=True)
```
evantey14, line 112: we should have a a more standard way to saying what defines an admin. just is_superuser? also is_staff? i guess we should be trying to keep them in sync? It would be nice if we could define some sort of is_admin property.

server/core/admin/utils.py
```diff
@@ -0,0 +1,135 @@
...
+            ("student", "Student"),
+            ("teacher", "Teacher"),
+            ("admin", "Admin"),
+        )
+
+    def queryset(self, request, queryset):
+        value = self.value()
+        if value == "student":
+            students = StudentProfile.objects.all().values_list("user", flat=True)
+            return queryset.filter(id__in=students)
```
evantey14, line 107: Doesn't have to be done now, but it 'd be nice if we got more in a habit of putting things like this under a simple named manager function e.g. queryset.get_all_students().

server/core/admin/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+    Class,
+    Program,
+    StudentClassRegistration,
+    StudentRegistration,
+    TeacherClassRegistration,
+    TeacherRegistration,
+)
+from django.contrib import admin
+
+from .users import *  # noqa
```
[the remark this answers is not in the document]
mvadari: Yeah, that's why I have it in there - nothing from `.users` is being used in the file, but the admin panel requires that in order to show everything that's in `.users`.

server/core/admin/utils.py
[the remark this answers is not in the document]
mvadari: Yeah fair, did that.

server/core/models/__init__.py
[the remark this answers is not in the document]
mvadari: I'll leave that as a TODO/card, since there are a bunch of these calls. 

Yeah, I did that for now because teacherreg hasn't been fleshed out. I was gonna add a similar filter for teacherreg. Honestly right now I don't have a better definition, unless we literally have a boolean called "active" that is turned off when a program goes to `POST_PROGRAM`.

server/core/admin/utils.py
mvadari: I added some more documentation to hopefully help - let me know if it needs more work.

server/core/admin/utils.py
mvadari: Oops done

server/core/admin/utils.py
mvadari: Yeah fair, I'll take it out.

server/core/admin/utils.py
mvadari: For now, I'm assuming that both `is_staff` and `is_superuser` are true if someone is an admin. I figured we would figure out that exact issue later, since we haven't really talked a whole lot about admins, and just went with this for now.

server/core/admin/utils.py
mvadari: We'll need to do a bit of an overhaul to introduce managers everywhere. Left a TODO

mvadari:
> I think in general, this is a bit long for a PR. Long PRs mean lots of comments and long merge times.

Yeah that's fair. It seemed like a lot of the same changes, just applied in a lot of places, so I didn't think much of it.

server/core/admin/__init__.py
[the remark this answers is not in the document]
evantey14: oh oh i see, yeah that sounds good

server/core/models/__init__.py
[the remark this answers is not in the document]
evantey14: yeah definitely fine for now, but we should leave a comment saying this is subject to change. Even now, I think it's already hard to tell which code is intentional and which code is just placeholder. I think this + people being afraid to change old code because they assume it was intentional is a quick recipe for a dirty codebase

server/core/admin/utils.py
evantey14: So yeah I think this is true, but I also think it's really important that we combine them soon (read: leave comment). If they're left separate and different parts check different properties, a reader might get confused. Also eventually we don't want to allow  them to have different values and the more code that only uses one makes it easier to mess that up.

server/core/models/__init__.py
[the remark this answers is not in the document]
mvadari: Yeah fair, left a note

server/core/admin/utils.py
evantey14: Yeah ok I think that helps me a bit. I think this is mostly an unfortunate side effect of Django...not sure the best way to make things more clear.

server/core/admin/utils.py
mvadari: Alright, left a comment. We might be able to add a constraint to `ESPUser` to ensure that these have the same value.

evantey14:
> > I think in general, this is a bit long for a PR. Long PRs mean lots of comments and long merge times.
> 
> Yeah that's fair. It seemed like a lot of the same changes, just applied in a lot of places, so I didn't think much of it.

Yeah so I think this is hopefully a good thing to get in the habit of doing. Even though my student signup form would be basically the same as teacher, I didn't do the teacher one because it's easier to do one, get the feedback, then finish the other instead of doing both, getting feedback, then changing both.

server/core/admin/utils.py
mvadari: changed wording a bit, maybe that helps?

mvadari:
> > > I think in general, this is a bit long for a PR. Long PRs mean lots of comments and long merge times.
> > 
> > 
> > Yeah that's fair. It seemed like a lot of the same changes, just applied in a lot of places, so I didn't think much of it.
> 
> Yeah so I think this is hopefully a good thing to get in the habit of doing. Even though my student signup form would be basically the same as teacher, I didn't do the teacher one because it's easier to do one, get the feedback, then finish the other instead of doing both, getting feedback, then changing both.

Alright, that makes a lot of sense. I'll make sure I do that.

PR #12, opened 2020-07-05
Incorporate new models

mvadari:
- Replace `clazz` with `section` in StudentClassRegistration
- Add Timeslot, Section, ScheduledBlock to admin panel
- Add sample data for Timeslot, Section, ScheduledBlock

Test: played around with the admin panel a bunch, ran the sample Python file and made sure everything looked imported correctly


server/core/admin/__init__.py
```diff
@@ -68,17 +72,46 @@ def get_search_results(self, request, queryset, search_term):
...
+    def get_search_results(self, request, queryset, search_term):
+        # this is a little bit hacky but checks if it's an autocomplete request from an Inline
+        # TODO figure out add
+        if "autocomplete" in request.path and "change" in request.META.get("HTTP_REFERER"):
+            section = int(request.META.get("HTTP_REFERER").split("/")[6])
+            if "section" in request.META.get("HTTP_REFERER"):
+                program = Section.objects.get(pk=section).clazz.program
+                queryset = queryset.filter(program__exact=program).order_by("start")
+
+        return super().get_search_results(request, queryset, search_term)
```
evantey14, line 109: hmm can we put this in a helper function? I'm very much not a fan of having this hacky code copied in so many places. Ideally we'd have a less hacky way to do it cause right now it's kinda opaque, but at the least we should make it so when we fix it, we don't have to fix it in like 5 locations.

server/core/models/__init__.py
```diff
@@ -146,6 +163,13 @@ class ScheduledBlock(models.Model):
     section = models.ForeignKey(Section, on_delete=models.CASCADE)
     timeslot = models.ForeignKey(Timeslot, on_delete=models.CASCADE)
 
+    class Meta:
+        unique_together = (("section", "timeslot"),)
+
```
evantey14, line 168: You should probably clear the TODO in the docstring.

server/core/admin/__init__.py
```diff
@@ -28,7 +35,7 @@ def save_model(self, request, obj, form, change):
 
     def add_view(self, request, extra_content=None):
         self.fields = (("name", "edition"),)
-        return super(ProgramAdmin, self).add_view(request, extra_content)
+        return super(ProgramAdmin, self).add_view(request)
```
evantey14, line 38: why did `extra_content` get removed?

server/core/models/__init__.py
```diff
@@ -93,6 +89,14 @@ class Section(models.Model):
...
+    @property
+    def program(self):
+        return self.clazz.program
+
+    @property
+    def num_students(self):
+        return StudentClassRegistration.objects.filter(section__id=self.id).count()
+
     def __str__(self):
         return str(self.clazz) + " sec. " + str(self.number)
```
evantey14, line 101: Hmm we should probably try to standardize the order of things inside classes. In StudentRegistration it looks like the toString is above the property functions? I don't think I particularly care for how we order things, but I think we should try pretty hard to have a consistent order? (This dictates how people add to code / find things)

server/core/admin/__init__.py
mvadari: The code has minor differences that makes it rather difficult to do that. For example, `program` obtaining happens differently in each function. I can make the `if` a helper function, but that's probably the best I can do.

server/core/models/__init__.py
mvadari: Oops, thought I did that. Done.

server/core/admin/__init__.py
mvadari: We discussed this during the meeting. Fixed.

server/core/models/__init__.py
mvadari: Hm I tried to do `Meta`-property-`__str__`-static methods but I guess I missed a couple - I think it was methods that got converted to property methods and not moved accordingly.

mvadari:
> Overall, looks good. I think this is still maybe 3 PRs masquerading as 1. (I could see adding stuff to admin panel as one, adding fake data as another, and adding the uniqueness constraints as a third)

yeahhhhhh I'll be Better.

server/core/admin/__init__.py
evantey14: Hmm yeah that's unfortunate. I guess we can leave it? but keep thinking if you can think of a way to make this more understandable.

mvadari:
Check out the documentation I put in the wiki - I can flesh that out more, but that should help.

PR #13, opened 2020-07-06
Better forms

mvadari:
Create better helper functions for forms so that the forms don't need to be as messy-looking. 

Test: went to all the pages with forms and checked that they rendered correctly

No change in what the forms look like, so I'm not attaching any screenshots


mvadari:
Yeah, fair. I figured this would at least abstract out those problems to the `helper` file, instead of the various files that currently have forms.

PR #14, opened 2020-07-07
Class catalog ui

mvadari:
- Set up class catalog API (basics, not fleshed out)
- First crack at class catalog UI

image

### DIFF
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -1,6 +1,7 @@
 import React, { Component } from "react";
 import axiosInstance from "../axiosAPI";
 import { Clazz } from "./types";
+import { studentScheduleEndpoint, classCatalogEndpoint } from "../apiEndpoints";
 
 type Props = {
   loggedIn: boolean;
@@ -10,8 +11,10 @@ type Props = {
 };
 
 type State = {
+  scheduleEndpoint: string;
+  catalogEndpoint: string;
   timeslots: string[];
-  classes: string[];
+  enrolledClasses: string[];
   catalog: Clazz[];
 };
 
@@ -34,35 +37,30 @@ class StudentRegDashboard extends Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
+      scheduleEndpoint: `/${this.props.program}/${this.props.edition}/${studentScheduleEndpoint}`,
+      catalogEndpoint: `/${this.props.program}/${this.props.edition}/${classCatalogEndpoint}`,
       timeslots: [],
-      classes: [],
+      enrolledClasses: [],
       catalog: [],
     };
   }
 
   componentDidMount() {
-    this.getStudentClasses();
-    this.getClassCatalog();
+    this.setupStudentClasses();
+    this.setupClassCatalog();
   }
 
-  getStudentClasses() {
-    axiosInstance
-      .get("/studentclasses/", {
-        params: {
-          program: this.props.program,
-          edition: this.props.edition,
-        },
-      })
-      .then(res => {
-        this.setState({
-          timeslots: res.data.timeslots,
-          classes: res.data.classes,
-        });
+  setupStudentClasses() {
+    axiosInstance.get(this.state.scheduleEndpoint).then(res => {
+      this.setState({
+        timeslots: res.data.timeslots,
+        enrolledClasses: res.data.classes,
       });
+    });
   }
 
-  getClassCatalog() {
-    axiosInstance.get(`/${this.props.program}/${this.props.edition}/catalog/`).then(res => {
+  setupClassCatalog() {
+    axiosInstance.get(this.state.catalogEndpoint).then(res => {
       this.setState({
         catalog: res.data,
       });
@@ -83,7 +81,7 @@ class StudentRegDashboard extends Component<Props, State> {
               </tr>
             </thead>
             <tbody>
-              {this.state.classes.map((clazz, index) => {
+              {this.state.enrolledClasses.map((clazz, index) => {
                 return (
                   <tr key={index}>
                     <th>{this.state.timeslots[index]}</th>
@@ -91,7 +89,7 @@ class StudentRegDashboard extends Component<Props, State> {
                     <td>
                       <button className="delete is-centered"></button>
                       {/*(for a) onClick=e => this.removeClass(e, clazz)
-                      TODO classes is Clazz[] instead of string[] */}
+                      TODO enrolledClasses is Clazz[] instead of string[] */}
                     </td>
                   </tr>
                 );
@@ -109,10 +107,12 @@ class StudentRegDashboard extends Component<Props, State> {
 
   addClass(e: React.MouseEvent, clazz: Clazz) {
     console.log("Adding " + clazz.title);
+    // TODO make this functional
   }
 
   removeClass(e: React.MouseEvent, clazz: Clazz) {
     console.log("Removing " + clazz.title);
+    // TODO make this functional
   }
 
   toggleClassDescription(e: React.MouseEvent) {
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -1,6 +1,6 @@
 import React, { Component } from "react";
 import axiosInstance from "../axiosAPI";
-import { Clazz } from "./types";
+import { Class } from "./types";
 import { studentScheduleEndpoint, classCatalogEndpoint } from "../apiEndpoints";
 
 type Props = {
@@ -11,11 +11,9 @@ type Props = {
 };
 
 type State = {
-  scheduleEndpoint: string;
-  catalogEndpoint: string;
-  timeslots: string[];
+  catalog: Class[];
   enrolledClasses: string[];
-  catalog: Clazz[];
+  timeslots: string[];
 };
 
 // helper functions
@@ -37,8 +35,6 @@ class StudentRegDashboard extends Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
-      scheduleEndpoint: `/${this.props.program}/${this.props.edition}/${studentScheduleEndpoint}`,
-      catalogEndpoint: `/${this.props.program}/${this.props.edition}/${classCatalogEndpoint}`,
       timeslots: [],
       enrolledClasses: [],
       catalog: [],
@@ -51,20 +47,24 @@ class StudentRegDashboard extends Component<Props, State> {
   }
 
   setupStudentClasses() {
-    axiosInstance.get(this.state.scheduleEndpoint).then(res => {
-      this.setState({
-        timeslots: res.data.timeslots,
-        enrolledClasses: res.data.classes,
+    axiosInstance
+      .get(`/${this.props.program}/${this.props.edition}/${studentScheduleEndpoint}`)
+      .then(res => {
+        this.setState({
+          timeslots: res.data.timeslots,
+          enrolledClasses: res.data.classes,
+        });
       });
-    });
   }
 
   setupClassCatalog() {
-    axiosInstance.get(this.state.catalogEndpoint).then(res => {
-      this.setState({
-        catalog: res.data,
+    axiosInstance
+      .get(`/${this.props.program}/${this.props.edition}/${classCatalogEndpoint}`)
+      .then(res => {
+        this.setState({
+          catalog: res.data,
+        });
       });
-    });
   }
 
   renderClassSchedule() {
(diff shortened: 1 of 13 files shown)


client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -57,6 +61,14 @@ class StudentRegDashboard extends Component<Props, State> {
       });
   }
 
+  getClassCatalog() {
```
evantey14, line 64: if we want this to have a side-effect we should start the function name with "setup". Alternatively, we can make it not have the side-effect and actually just return the catalog (and do the setState in componentDidMount or whereever this function is called).

(this is also true of the other setup function)

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -85,12 +103,63 @@ class StudentRegDashboard extends Component<Props, State> {
     );
   }
 
+  enrolledInClass(clazz: Clazz) {
+    return false;
+  }
```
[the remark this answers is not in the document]
mvadari: I thought it would check if the class was in the list of classes that the student was enrolled in. Should I do that in the backend or the frontend?

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -85,12 +103,63 @@ class StudentRegDashboard extends Component<Props, State> {
...
+
+  addClass(e: React.MouseEvent, clazz: Clazz) {
+    console.log("Adding " + clazz.title);
+  }
+
+  removeClass(e: React.MouseEvent, clazz: Clazz) {
+    console.log("Removing " + clazz.title);
+  }
```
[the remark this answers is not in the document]
mvadari: Done

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -85,12 +103,63 @@ class StudentRegDashboard extends Component<Props, State> {
...
+
+  toggleClassDescription(e: React.MouseEvent) {
+    e!.currentTarget!.parentElement!.nextElementSibling!.classList.toggle("is-hidden");
```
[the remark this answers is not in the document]
mvadari: As far as I understand, the exclamation point is a guard against null values that may arise. It doesn't compile without all the exclamation points.

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -11,6 +12,7 @@ type Props = {
 type State = {
   timeslots: string[];
   classes: string[];
+  catalog: Clazz[];
```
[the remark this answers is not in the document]
mvadari: > I think it's a bit confusing at first why classes and catalog are different?
This is temporary, it was easier to generate a sample API output with string names rather than the whole class object. 

> I think it'd be nice if we could name classes something that makes it clear "classes" are the student's classes (the classes the student are enrolled in?)
Renamed to `enrolledClasses`

> Somewhat relatedly, I feel like I kinda want better api endpoints for these. Something like "/{prog}/classes" and "/{prog}/{user}/classes"? (I feel like I don't know the right way to do this, but it doesn't quite sit right if studentclasses has params but those params are included in the route for catalog?)
Whoops, forgot to update the API endpoints. I split the work I did into 2 PRs (this is the first) so there's some stuff that didn't transfer correctly.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: Renamed the function (and all analogous functions elsewhere).

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -9,8 +11,11 @@ type Props = {
 };
 
 type State = {
+  scheduleEndpoint: string;
+  catalogEndpoint: string;
```
wishdasher, line 15: If `program` and `edition` are props and the endpoint string isn't changing, there is no need for those two to be part of the state. The strings can be functionally determined from the props.

client/src/registration/ChangeClassesDashboard.tsx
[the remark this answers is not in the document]
wishdasher: Yikes. There are better ways to set up a toggle than using DOM manipulation, but I would leave a TODO for now to remove this later.

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -85,12 +101,65 @@ class StudentRegDashboard extends Component<Props, State> {
...
+  renderClass(clazz: Clazz) {
+    return (
+      <div className="card" key={clazz.id}>
+        <header className="card-header">
+          <h2 className="card-header-title">{clazz.title}</h2>
+          <a
+            href="# "
+            className="card-header-icon card-toggle"
+            onClick={this.toggleClassDescription}
+          >
```
wishdasher, line 131: We should not have empty links, bad for screen readers. Did React not warm you about this?

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -85,12 +101,65 @@ class StudentRegDashboard extends Component<Props, State> {
...
+  }
+
+  toggleClassDescription(e: React.MouseEvent) {
+    e!.currentTarget!.parentElement!.nextElementSibling!.classList.toggle("is-hidden");
+  }
+
+  renderClass(clazz: Clazz) {
+    return (
+      <div className="card" key={clazz.id}>
+        <header className="card-header">
```
wishdasher, line 125: Maybe the docs told you to use `header`, but I would like to reserve `header` and `footer` (below) for page-level header and footer. This [page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header) says the implicit aria role is a banner, which I don't think applies to this use case.

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -85,12 +101,65 @@ class StudentRegDashboard extends Component<Props, State> {
...
+        </header>
+        <div className="card-content is-hidden">
+          <div className="content">
+            <p>
+              <i>{clazz.teachers.join(", ")}</i>
+            </p>
+            <p>{clazz.description}</p>
+          </div>
+        </div>
+        <footer className="card-footer">
```
wishdasher, line 145: See above: `div` should work fine for these.

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -85,12 +101,65 @@ class StudentRegDashboard extends Component<Props, State> {
...
+  }
+
   renderClassCatalog() {
     return (
       <div className="column">
         <h2 className="has-text-centered is-size-3">Class Catalog</h2>
-        <br />
-        {renderTextInSection("Class catalog here", false)}
+        {renderTextInSection("Insert filter options here")}
+        {this.state.catalog.map((clazz, index) => this.renderClass(clazz))}
```
wishdasher, line 162: If you're not using the `index`, you can leave it out e.g.
`{this.state.catalog.map(class => ...)}`

client/src/registration/StudentRegDashboard.tsx
```diff
@@ -11,6 +11,8 @@ type Props = {
 };
 
 type State = {
+  regEndpoint: string;
+  scheduleEndpoint: string;
```
wishdasher, line 15: Same comment about these not being part of the state.

client/src/registration/types.ts
```diff
@@ -1,6 +1,15 @@
 // should match RegStatusOptions in the backend
 // TODO figure out how to sync them
 
+type Clazz = {
+  id: string;
+  title: string;
+  description: string;
+  teachers: string[];
+  capacity: number;
+  num_students: number;
```
wishdasher, line 10: There may be some semantic ordering here, but I think we could at least have: `id`, `title`, `description`, and then the rest being alphabetized.

server/core/views/api.py
```diff
@@ -61,6 +62,18 @@ class TeacherProgramViewSet(viewsets.ReadOnlyModelViewSet):
     serializer_class = ProgramSerializer
 
 
+class ClassCatalog(APIView):
+    """
+    API endpoint that returns all classes.
+    Permissions: authenticated (should be anyone eventually)
```
wishdasher, line 68: Is this last comment a TODO? (Add in a `TODO`.)

client/src/registration/ChangeClassesDashboard.tsx
wishdasher: An alternative is using `button`, but that usually come with awkward style defaults. I tried making an unstyled one for the nav, look in that css file.

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -1,5 +1,7 @@
 import React, { Component } from "react";
 import axiosInstance from "../axiosAPI";
+import { Clazz } from "./types";
```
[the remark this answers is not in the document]
mvadari: Oh rad, done

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -9,8 +11,11 @@ type Props = {
 };
 
 type State = {
+  scheduleEndpoint: string;
+  catalogEndpoint: string;
   timeslots: string[];
-  classes: string[];
+  enrolledClasses: string[];
+  catalog: Clazz[];
```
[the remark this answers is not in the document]
mvadari: Done

client/src/registration/ChangeClassesDashboard.tsx
mvadari: Ok, done

client/src/registration/ChangeClassesDashboard.tsx
mvadari: StackOverflow said that empty links were fine with the "# ". I can still replace this with a button if you think that's better, though.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: Alright, I think I literally just copied from the docs. Fixed.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: Done

client/src/registration/StudentRegDashboard.tsx
mvadari: Done

client/src/registration/types.ts
mvadari: Done

server/core/views/api.py
mvadari: Done

client/src/registration/ChangeClassesDashboard.tsx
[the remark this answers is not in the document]
evantey14: oh ok right because there are enrolled classes and the full list of classes. Either is probably fine -- maybe add in a TODO explaining that this is gonna compare stuff in the state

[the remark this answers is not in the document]
mvadari: Done.

client/src/registration/ChangeClassesDashboard.tsx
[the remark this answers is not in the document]
mvadari: Done.

client/src/registration/ChangeClassesDashboard.tsx
wishdasher: Okay, I say keep it as a link, add a `role="button"`, and maybe do a `href="#toggleDescription"` or `href="#void"` instead. I'm worried about someone removing that space after the # in the future, and that would have bad default behavior.

wishdasher:
Just that one last comment about the empty anchor tag.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: Done, added a `preventDefault` in the function as well so it doesn't refresh the page.

wishdasher: Buttons and anchors have become very conflated in modern web dev. You have links that look like buttons, buttons that perform as links...I digress. The major downside to this solution that I can think of is that I would expect, by looking at this element (which acts like a button), that it could be triggered with a space press. Unfortunately, not true for anchors.

mvadari: Honestly for toggles I think a space press is reasonable - if I'm on a button/anchor that I expect to toggle something, then I would expect that a space press does that. But maybe that's because everyone uses anchors and buttons.

PR #15, opened 2020-07-08
Implement most model constraints

mvadari:
- Some done as Django CheckConstraints
  - Some done on __init__ of the object
  - Some done on clean() in save()

Test: Played around with all the models in the admin panel, made sure an error was raised when violating one or more of the imposed constraints


evantey14:
Three feedback points/questions about code:
* What kind of error message do the `models.checkConstraint` things give? Can we control the message? Are they informative enough?
* For your Validation Errors, we should use parameters to be more explicit about the error. e.g. when we check for program consistency, we should say which object got checked, and which programs were in disagreement. Doing this gives us more information about what's going on. (Imagine just seeing "programs aren't the same" pop up as part of some more complicated bulk action like the lottery -- this wouldn't really tell you what happened / how to track it down / how to fix it).
* I am mildly worried about validation being scattered in too many places (save vs constraints property vs clean). I haven't looked into it much so what you've done may well be the right way to organize it, but just something to be mindful about

wishdasher:
I should have explicitly mentioned this at a meeting, but I intentionally didn't implement constraints for the models I added because they add limitations without obviously facilitating progress, and the system should be thought through holistically at some point anyways.

mvadari:
Alright, that's fair. 

I figure we should at least have this code somewhere for reference later, when we do implement constraints - should I just close the PR?

evantey14:
Yep, let's close it for now.

PR #16, opened 2020-07-09
Make class catalog more functional

mvadari:
More additions to the class catalog

* Functional API for schedule retrieval
* Make the delete button in the schedule use the placeholder `removeClass` function
* Add placeholder search/filtering options

image

### DIFF
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -1,6 +1,6 @@
 import React, { Component } from "react";
 import axiosInstance from "../axiosAPI";
-import { Class, ScheduledTimeslot } from "./types";
+import { Class, Section, ScheduledTimeslot } from "./types";
 import { studentScheduleEndpoint, classCatalogEndpoint } from "../apiEndpoints";
 import { renderCustomInput } from "../forms/helpers";
 
@@ -88,11 +88,11 @@ class StudentRegDashboard extends Component<Props, State> {
                 return (
                   <tr key={index}>
                     <th>{scheduleItem.timeslot}</th>
-                    <td>{scheduleItem.clazz && scheduleItem.clazz.title}</td>
+                    <td>{scheduleItem.section && scheduleItem.section.name}</td>
                     <td>
-                      {scheduleItem.clazz && (
+                      {scheduleItem.section && (
                         <button
-                          onClick={e => this.removeClass(e, scheduleItem.clazz)}
+                          onClick={e => this.removeClass(e, scheduleItem.section)}
                           className="delete is-centered"
                         ></button>
                       )}
@@ -125,10 +125,10 @@ class StudentRegDashboard extends Component<Props, State> {
     // TODO make this functional
   }
 
-  removeClass(e: React.MouseEvent, clazz: Class) {
+  removeClass(e: React.MouseEvent, section: Section) {
     e.preventDefault();
-    if (clazz) {
-      console.log("Removing " + clazz.title);
+    if (section) {
+      console.log("Removing " + section.name);
       // TODO make this functional
     }
   }
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -14,7 +14,6 @@ type State = {
   catalog: Class[];
   enrolledClasses: string[];
   timeslots: string[];
-  catalog: Clazz[];
 };
 
 // helper functions
@@ -116,11 +115,13 @@ class StudentRegDashboard extends Component<Props, State> {
   }
 
   addClass(e: React.MouseEvent, clazz: Class) {
+    e.preventDefault();
     console.log("Adding " + clazz.title);
     // TODO make this functional
   }
 
   removeClass(e: React.MouseEvent, clazz: Class) {
+    e.preventDefault();
     console.log("Removing " + clazz.title);
     // TODO make this functional
   }
@@ -135,7 +136,9 @@ class StudentRegDashboard extends Component<Props, State> {
     return (
       <div className="card" key={clazz.id}>
         <div className="card-header">
-          <h2 className="card-header-title">{clazz.title}</h2>
+          <a href="# " className="card-header-title" onClick={this.toggleClassDescription}>
+            {clazz.title}
+          </a>
           <a
             href="#void"
             role="button"
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -1,6 +1,6 @@
 import React, { Component } from "react";
 import axiosInstance from "../axiosAPI";
-import { Class } from "./types";
+import { Class, ScheduledTimeslot } from "./types";
 import { studentScheduleEndpoint, classCatalogEndpoint } from "../apiEndpoints";
 import { renderCustomInput } from "../forms/helpers";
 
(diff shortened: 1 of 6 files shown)


client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -81,22 +84,26 @@ class StudentRegDashboard extends Component<Props, State> {
               </tr>
             </thead>
             <tbody>
-              {this.state.enrolledClasses.map((clazz, index) => {
+              {this.state.schedule.map((scheduleItem, index) => {
```
evantey14, line 87: nit: I think if you're not using index you can just not include it?

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -156,18 +175,71 @@ class StudentRegDashboard extends Component<Props, State> {
             >
               Add Class
             </a>
-          }
-          <h3 className="card-footer-item">{`${clazz.capacity} students`}</h3>
+          ) : (
+            <a href="#void" className="card-footer-item" role="button">
+              Join waitlist
```
evantey14, line 180: this is a great feature! let's add a todo somewhere (if you haven't already)

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -147,7 +166,7 @@ class StudentRegDashboard extends Component<Props, State> {
           </div>
         </div>
         <div className="card-footer">
-          {
+          {classHasSpace ? (
```
evantey14, line 169: Is a large ternary operator the right way to do this? (probably leaving for @wishdasher to comment, but I feel like this is a bit hard to read.)

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -156,18 +175,71 @@ class StudentRegDashboard extends Component<Props, State> {
...
   }
 
+  handleSearchChange = (e: React.FormEvent<HTMLInputElement>) => {
+    this.setState({
+      search: e.currentTarget.value,
+    });
+  };
+
+  submitSearch = (e: React.MouseEvent<HTMLButtonElement>) => {
+    console.log("Searching for " + this.state.search);
```
evantey14, line 200: nit: TODO? Is the plan to make this an actual request? or just do something on the frontend?

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -156,18 +175,71 @@ class StudentRegDashboard extends Component<Props, State> {
...
+          </h3>
         </div>
       </div>
     );
   }
 
+  handleSearchChange = (e: React.FormEvent<HTMLInputElement>) => {
+    this.setState({
+      search: e.currentTarget.value,
+    });
```
evantey14, line 196: If this will eventually be used just to make a request, do we need this to be stored in the state? Or can we just do it old-school style.

(Also if we do keep it in state, I'd like a better name like classQuery or something)

client/src/registration/types.ts
```diff
@@ -20,6 +10,29 @@ enum RegStatusOption {
...
+  num_students: number;
+  number: number;
+};
+
+type Class = {
+  id: string;
+  title: string;
+  description: string;
+  capacity: number;
+  section_set: Section[];
```
evantey14, line 29: Is there a reason this is section_set and not sections? I feel like set has implied data structure connotations that aren't true here?

server/core/models/__init__.py
```diff
@@ -107,6 +111,10 @@ def program(self):
     def num_students(self):
         return StudentClassRegistration.objects.filter(section__id=self.id).count()
 
+    @property
+    def scheduledblocks(self):
```
evantey14, line 115: nit: snake case?

client/src/registration/types.ts
```diff
@@ -20,6 +10,29 @@ enum RegStatusOption {
   Empty = "",
 }
 
+
+type ScheduledTimeslot = {
+  timeslot: string;
+  clazz: Class;
```
evantey14, line 16: should this be section? (I think this also has many implications for the rest of the changes in this PR)

mvadari:
> I think even at this point it might be a good move to try taking those last 2 out into new PRs. 

How do I do this? I wanted to separate it out, but the commits are in the middle of everything (should have done it in a separate branch), and I'm not good enough at git to determine how to do it.

server/core/models/__init__.py
mvadari: I don't think this should be snake case because the object is called `ScheduledBlock`

client/src/registration/ChangeClassesDashboard.tsx
mvadari: `index` is used for the `tr` key.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: Done.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: I think it'll be easier to do this with an API call rather than on the frontend, given Django's search capabilities. I added a TODO.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: This needs to be stored in the state so the textbox shows what you've typed so far, just like all the other forms.

mvadari: I renamed it though.

client/src/registration/types.ts
mvadari: That's how Django auto-creates this - `{model_name}_set`. The models need to be mildly changed for that to happen.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: Honestly I don't know, but it was the best way I could think of doing this.

server/core/models/__init__.py
wishdasher: Is that how it works? The `B` signifies a new word, I don't think I have seen too much of preservation of atomicity of object names, but that could be a thing I suppose.

client/src/registration/types.ts
mvadari: Class made sense to me because it's class information that you would want on this page (e.g. title, description).

mvadari: Replaced it, wasn't too complicated.

client/src/registration/ChangeClassesDashboard.tsx
evantey14: Oh lol you're right.

client/src/registration/types.ts
evantey14: I think we should be able to use a `related_name` property in the field definition? I'd strongly prefer that because section_set is no longer really a set by the time it gets to the frontend...

evantey14:
> > I think even at this point it might be a good move to try taking those last 2 out into new PRs.
> 
> How do I do this? I wanted to separate it out, but the commits are in the middle of everything (should have done it in a separate branch), and I'm not good enough at git to determine how to do it.

Yeah so I think the right command is cherry-pick: https://git-scm.com/docs/git-cherry-pick

You'd want to create new branches (one for each feature) and cherry pick relevant commits from class-catalog-2 into the new branches. This shouldn't be too dangerous since you're just reading from class-catalog-2 (shouldn't be at risk of losing work).

mvadari:
Closing this PR to split it into multiple PRs based on cherry-picking.

PR #18, opened 2020-07-11
Remove Django templates, views, and forms

evantey14:
Remove all Django template frontend stuff (and connected views / forms) since we now have a large amount of it in React.

All I did for testing was derp around the React frontend (e.g. went through the student reg process up until entering class prefs).


evantey14:
Also @mvadari I had a question about "/api/dashboard/teacher". It doesn't seem to work (but I don't think my changes broke it)? Was this already broken?

mvadari:
> Also @mvadari I had a question about "/api/dashboard/teacher". It doesn't seem to work (but I don't think my changes broke it)? Was this already broken?

Yeah, I'm on master and it's not working. The teacher dashboard needs to be reworked anyways eventually, so it's not a huge deal.

PR #19, opened 2020-07-11
Fix trailing slash handling

evantey14:
Right now, we're catching all url patterns and sending them to React after one pass through the urlpatterns list. This means Django's trailing slash handling doesn't work (if a url doesn't have a trailing slash, it first makes one pass through urlpatterns, and if it doesn't match anything Django appends a slash and makes a second pass).

This is fixed if our React wildcard pattern requires a trailing slash at the end.

Testing: Lazily tested by checking localhost:8000/admin works.

Draft because I want to wait on #18 to get merged first.


mvadari:
Did you test with the React paths and API paths?

evantey14:
> Did you test with the React paths and API paths?

Just tested this -- looks like navigating student dashboard and stuff appears to work. Also able to hit API endpoints (e.g.  "api/token") both with and without slashes. 

This is ready for review now.

PR #20, opened 2020-07-11
Add class catalog filtering option placeholders

mvadari:
* Added search bar 
* Added "only open classes" toggle select

Test: 
* typed things into the search bar, hit "search", checked logs
* hit the "only open classes" toggle a bunch, checked logs

image

### DIFF
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -2,6 +2,7 @@ import React, { Component } from "react";
 import axiosInstance from "../axiosAPI";
 import { Class } from "./types";
 import { studentScheduleEndpoint, classCatalogEndpoint } from "../apiEndpoints";
+import { renderCustomInput } from "../forms/helpers";
 
 type Props = {
   loggedIn: boolean;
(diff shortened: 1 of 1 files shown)


client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -13,15 +14,14 @@ type Props = {
 type State = {
   catalog: Class[];
   enrolledClasses: string[];
+  openClasses: boolean;
```
evantey14, line 17: I think I'd prefer if this were something about filtering or displaying  open classes? From this name, I would expect a Class array, not a boolean

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -13,15 +14,14 @@ type Props = {
...
+  search: string;
```
evantey14, line 18: same deal earlier about a more descriptive name than just "search" I think even "searchQuery" would be better / make more sense (maybe this is just how I read it, but to me, search is more a verb than a noun?)

client/src/registration/ChangeClassesDashboard.tsx
mvadari: Done

client/src/registration/ChangeClassesDashboard.tsx
mvadari: done

PR #22, opened 2020-07-11
Display actual student schedule

mvadari:
* Make a functional API for retrieving a student's schedule and displaying it in the schedule. 
* Adjust frontend accordingly, so it displays properly.

image

Test:
* Gave the student various sets of classes, made sure they rendered properly

Assumes you can't have 2 classes scheduled at the same time (for which there is no current check).

### DIFF
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -16,7 +16,6 @@ type State = {
   displayOnlyOpenClasses: boolean;
   enrolledClasses: string[];
   timeslots: string[];
-  catalog: Clazz[];
 };
 
 // helper functions
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -16,6 +16,7 @@ type State = {
   displayOnlyOpenClasses: boolean;
   enrolledClasses: string[];
   timeslots: string[];
+  catalog: Clazz[];
 };
 
 // helper functions
@@ -66,6 +67,14 @@ class ClassChangesDashboard extends Component<Props, State> {
           catalog: res.data,
         });
       });
+  }
+
+  getClassCatalog() {
+    axiosInstance.get(`/${this.props.program}/${this.props.edition}/catalog/`).then(res => {
+      this.setState({
+        catalog: res.data,
+      });
+    });
   }
 
   renderClassSchedule() {
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -68,13 +68,6 @@ class ClassChangesDashboard extends Component<Props, State> {
       });
   }
 
-  getClassCatalog() {
-    axiosInstance.get(`/${this.props.program}/${this.props.edition}/catalog/`).then(res => {
-      this.setState({
-        catalog: res.data,
-      });
-    });
-  }
 
   renderClassSchedule() {
     return (
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -114,6 +114,12 @@ class ClassChangesDashboard extends Component<Props, State> {
     // TODO make this functional
   }
 
+  addWaitlistClass(e: React.MouseEvent, clazz: Class) {
+    e.preventDefault();
+    console.log("Adding to waitlist " + clazz.title);
+    // TODO make this functional
+  }
+
   removeClass(e: React.MouseEvent, clazz: Class) {
     e.preventDefault(); // TODO use button instead of anchor so this isn't needed
     if (clazz) {
@@ -173,7 +179,12 @@ class ClassChangesDashboard extends Component<Props, State> {
               Add Class
             </a>
           ) : (
-            <a href="#void" className="card-footer-item" role="button">
+            <a
+              href="#void"
+              className="card-footer-item"
+              role="button"
+              onClick={e => this.addWaitlistClass(e, clazz)}
+            >
               Join waitlist
             </a>
           )}
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -1,6 +1,6 @@
 import React, { Component } from "react";
 import axiosInstance from "../axiosAPI";
-import { Class, Section, ScheduledTimeslot } from "./types";
+import { Class, Section, ScheduleItem } from "./types";
 import { studentScheduleEndpoint, classCatalogEndpoint } from "../apiEndpoints";
 import { renderCustomInput } from "../forms/helpers";
 
@@ -14,7 +14,7 @@ type State = {
   catalog: Class[];
   classSearchQuery: string;
   displayOnlyOpenClasses: boolean;
-  schedule: ScheduledTimeslot[];
+  schedule: ScheduleItem[];
 };
 
 // helper functions
@@ -47,7 +47,11 @@ class ClassChangesDashboard extends Component<Props, State> {
 
   setupStudentClasses() {
     axiosInstance
-      .get(`/${this.props.program}/${this.props.edition}/${studentScheduleEndpoint}`)
+      .get(`/${this.props.program}/${this.props.edition}/${studentScheduleEndpoint}`, {
+        params: {
+          include_empty_timeslots: true,
+        },
+      })
       .then(res => {
         this.setState({
           schedule: res.data,
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -1,6 +1,6 @@
 import React, { Component } from "react";
 import axiosInstance from "../axiosAPI";
-import { Class } from "./types";
+import { Class, ScheduledTimeslot } from "./types";
 import { studentScheduleEndpoint, classCatalogEndpoint } from "../apiEndpoints";
 import { renderCustomInput } from "../forms/helpers";
 
@@ -14,8 +14,7 @@ type State = {
   catalog: Class[];
   classSearchQuery: string;
   displayOnlyOpenClasses: boolean;
-  enrolledClasses: string[];
-  timeslots: string[];
+  schedule: ScheduledTimeslot[];
 };
 
 // helper functions
@@ -34,11 +33,10 @@ class ClassChangesDashboard extends Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
-      catalog: [],
       classSearchQuery: "",
+      catalog: [],
       displayOnlyOpenClasses: true,
-      enrolledClasses: [],
-      timeslots: [],
+      schedule: [],
     };
   }
 
@@ -52,8 +50,7 @@ class ClassChangesDashboard extends Component<Props, State> {
       .get(`/${this.props.program}/${this.props.edition}/${studentScheduleEndpoint}`)
       .then(res => {
         this.setState({
-          timeslots: res.data.timeslots,
-          enrolledClasses: res.data.classes,
+          schedule: res.data,
         });
       });
   }
@@ -82,11 +79,11 @@ class ClassChangesDashboard extends Component<Props, State> {
               </tr>
             </thead>
             <tbody>
-              {this.state.enrolledClasses.map((clazz, index) => {
+              {this.state.schedule.map((scheduleItem, index) => {
                 return (
                   <tr key={index}>
-                    <th>{this.state.timeslots[index]}</th>
-                    <td>{clazz}</td>
+                    <th>{scheduleItem.timeslot}</th>
+                    <td>{scheduleItem.clazz && scheduleItem.clazz.title}</td>
                     <td>
                       <button className="delete is-centered"></button>
                       {/*(for a) onClick=e => this.removeClass(e, clazz)
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -133,7 +133,7 @@ class ClassChangesDashboard extends Component<Props, State> {
     // TODO replace this with something that actually checks this once sections have been
     // properly implemented
     return (
-      <div className="card" key={clazz.id}>
+      <div className="card mb-1" key={clazz.id}>
         <div className="card-header">
           <a
             href="#void"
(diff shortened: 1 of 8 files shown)


server/core/models/registration.py
```diff
@@ -31,19 +31,33 @@ class StudentRegistration(models.Model):
...
+    def get_schedule(self):
+        schedule = {timeslot: None for timeslot in self.program.timeslots}
+        sections = self.sections
+        for section in sections:
+            for block in section.scheduledblock_set.all():
+                timeslot = block.timeslot
+                schedule[timeslot] = section
+        # TODO add check to make sure sections don't overlap timeslots
+        # (or just make sure that's not possible when adding)
+        return sorted(list(schedule.items()), key=lambda pair: pair[0].start)
```
evantey14, line 59: So it looks like you chose to make a schedule all the program timeslots along with the student's section or None if they don't have one. Was there a particular reason for this (instead of just only having timeslots/sections where a student has classes?) Haven't really thought about it much, but I think my first instinct is to not include a bunch of None sections?

client/src/registration/types.ts
```diff
@@ -10,7 +10,14 @@ enum RegStatusOption {
   Empty = "",
 }
 
+type ScheduledTimeslot = {
```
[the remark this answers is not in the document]
mvadari: `ScheduledSection` didn't make sense to me, because multi-hour sections are being split up here by timeslot. What about something like `ScheduleItem`?

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -35,11 +34,10 @@ class StudentRegDashboard extends Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
-      catalog: [],
       classSearchQuery: "",
+      catalog: [],
```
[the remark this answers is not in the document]
mvadari: Unintentional, fixed.

server/core/models/program.py
```diff
@@ -62,16 +66,25 @@ class Timeslot(models.Model):
     end = models.DateTimeField()
     program = models.ForeignKey(Program, on_delete=models.CASCADE)
 
+    @property
+    def shortstr(self):
```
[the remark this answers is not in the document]
mvadari: what about something like `displaystr`? The idea is that this is the small string that is displayed in the UI in things like the schedule and the catalog.

server/core/models/registration.py
```diff
@@ -31,19 +31,33 @@ class StudentRegistration(models.Model):
...
     class Meta:
         unique_together = (("student", "program"),)
 
     def __str__(self):
         return str(self.student.username) + "/" + str(self.program)
 
+    def get_schedule(self):
+        schedule = {timeslot: None for timeslot in self.program.timeslots}
+        sections = self.sections
+        for section in sections:
```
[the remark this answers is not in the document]
mvadari: (y)

server/core/models/registration.py
mvadari: My thought was that if you're looking at your schedule in a pre-program stage, then you'll want to know what timeslots are empty, so that you know what times to look at if you want to add classes. 

Maybe there could be a parameter that asks whether to show the empty timeslots or not?

evantey14: So I think it's certainly needed, but I think the question is whether or those empty sections belong in this datatype we're starting to define as a schedule. Semantically, I'm not sure I would consider (11 AM, None) as part of a student's schedule (or at least that's my first instinct)? We want this instinct to match our API as much as possible so we can minimize confusion / misuse / people needing to refer to documentation.

The alternative would be to have something on the frontend to insert those empty sections after receiving a student's schedule.

client/src/registration/types.ts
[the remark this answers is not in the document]
evantey14: Mmm yeah that's a good point. That name seems better. I guess if we're trying to be consistent about language, maybe scheduledblock is the right term?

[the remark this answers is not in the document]
mvadari: I thought about that but that's not quite accurate, since it's not actually really a scheduled block. Should it be reformatted to be that?

server/core/models/registration.py
mvadari: I do think it's a useful functionality to include, in a lot of cases. What about some sort of boolean that defaults to false, called like `include_empty_timeslots`?

client/src/registration/types.ts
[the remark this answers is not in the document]
evantey14: Yeah I agree that it's not actually a scheduled block, (though also if we had called it a scheduled event, then this would've made slightly more sense). I think at this point a scheduleitem seems fine. As per the discussion we had at worksession, we may eventually want to be able to add ceremonies/meal (but we don't have to do that now)

server/core/models/program.py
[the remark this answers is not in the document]
evantey14: I think `displaystr` suffers similar problems. Seeing `timeslot.displaystr` doesn't tell you what gets displayed. Someone would have to go find this property, at which point they would see the format strings, at which point they would have to look up how those format strings work.

I think I'd prefer a title that says what's getting displayed like `start_time_str` or `start_end_times_str`? Thoughts?

server/core/models/registration.py
```diff
@@ -39,19 +40,34 @@ class StudentRegistration(models.Model):
     availability_check = models.BooleanField(default=False)
     payment_check = models.BooleanField(default=False)
 
+    @property
+    def sections(self):
+        sections = esp_models.StudentClassRegistration.objects.filter(studentreg=self).values_list(
+            "section", flat=True
+        )
+        return Section.objects.filter(pk__in=sections)
```
evantey14, line 48: Can we do this with related_names now that it got merged?

server/core/models/registration.py
```diff
@@ -39,19 +40,34 @@ class StudentRegistration(models.Model):
...
+
     @property
     def classes(self):
-        sections = self.classregs.values_list("section", flat=True)
-        ids = Section.objects.filter(pk__in=sections).values_list("clazz", flat=True)
+        ids = self.sections.values_list("clazz", flat=True)
         return Class.objects.filter(id__in=ids)
```
evantey14, line 53: Are we able to do this with something like `self.sections.classes`?

server/core/models/registration.py
evantey14: I think that's probably a fine solution

server/core/serializers.py
```diff
@@ -58,10 +66,14 @@ class Meta:
...
     scheduled_blocks = serializers.StringRelatedField(many=True)
+    name = serializers.SerializerMethodField()
 
     class Meta:
         model = Section
-        fields = ("clazz", "number", "num_students", "scheduled_blocks")
+        fields = ("clazz", "name", "number", "num_students", "scheduled_blocks")
+
+    def get_name(self, instance):
+        return str(instance)
```
evantey14, line 76: nit: would it be good to use `section` instead of `instance`?

server/core/views/studentreg.py
```diff
@@ -244,3 +242,4 @@ def get_object(self, program, edition):
         studentreg = StudentRegistration.objects.get(student=user, program=program)
 
         return studentreg
+
```
evantey14, line 245: nit: remove?

client/src/registration/types.ts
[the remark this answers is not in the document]
mvadari: Renamed.

shardulc:
Continuing the discussion about empty timeslots in schedules, here's a different idea. I think a schedule contains events, some of which are classes—and you cannot be in more than one class at any given time—and some of which are mealtimes, opening/closing ceremonies, maybe other things, and these *can* theoretically overlap with other events. So maybe the schedule API should return a list of events, each of which has a start and end time, and then the client side determines how to display the schedule and deal with overlaps and empty slots.

A side benefit of this is that it might become easier to deal with non-whole-hour times throughout the code in this perspective. For example, Summer HSSP this year has half-hour classes and an opening ceremony that started at 11:15. Actually I'm thinking back to the earlier discussion about pulling in components when you can instead of writing them yourself: would it be easier to just use the Google Calendar API or something for all this?

server/core/models/program.py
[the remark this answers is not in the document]
mvadari: hat about `date_time_str`?

server/core/models/registration.py
mvadari: Done.

server/core/models/registration.py
mvadari: Done.

server/core/serializers.py
mvadari: Yeah that works. I left it as `instance` initially because this is like a specific format, but it doesn't seem to complain about renaming parameters so it's fine.

server/core/views/studentreg.py
mvadari: Done.

mvadari:
> So maybe the schedule API should return a list of events, each of which has a start and end time, and then the client side determines how to display the schedule and deal with overlaps and empty slots.

I feel like that would be some sort of `events` API, not a schedule one. "Schedule" to me implies that time is the main factor. 

> A side benefit of this is that it might become easier to deal with non-whole-hour times throughout the code in this perspective.

Everything we've been doing has been assuming any (reasonable) timeblock size. 

> Actually I'm thinking back to the earlier discussion about pulling in components when you can instead of writing them yourself: would it be easier to just use the Google Calendar API or something for all this?

I don't quite understand where the Google Calendar API comes from. Are you talking about just displaying a Google Calendar on the webpage?

shardulc:
> > Actually I'm thinking back to the earlier discussion about pulling in components when you can instead of writing them yourself: would it be easier to just use the Google Calendar API or something for all this?
> 
> I don't quite understand where the Google Calendar API comes from. Are you talking about just displaying a Google Calendar on the webpage? I think that would be reasonable to consider down the line, but isn't necessary to consider now - right now we're focusing on the ESP logic-specific parts of the website, and you don't need any ESP knowledge to convert set of events and timeslots to a Google Calendar view.

Yeah, let me clarify. What I meant is that if we find ourselves writing our own implementation of concepts like 'event with start and end time' and actions like 'return a view on all events with property X', then we could consider using an existing calendaring API like Google Calendar to do this for us. So the basic entity of a 'class' would exist in GCal and the ESP-specific relations between a class and a student, etc. would exist in our website. Additional ESP-specific properties for these GCal events could be specified with something like [private extended properties](https://developers.google.com/calendar/extended-properties).

I don't know if this is actually a good idea at all. Maybe the only thing it makes easier is the front end. :woman_shrugging:

mvadari:
I think this is unnecessary. We're doing pretty good with our own objects, and I'd rather avoid bringing in other APIs for core parts of this website.

evantey14:
> > > Actually I'm thinking back to the earlier discussion about pulling in components when you can instead of writing them yourself: would it be easier to just use the Google Calendar API or something for all this?
> > 
> > 
> > I don't quite understand where the Google Calendar API comes from. Are you talking about just displaying a Google Calendar on the webpage? I think that would be reasonable to consider down the line, but isn't necessary to consider now - right now we're focusing on the ESP logic-specific parts of the website, and you don't need any ESP knowledge to convert set of events and timeslots to a Google Calendar view.
> 
> Yeah, let me clarify. What I meant is that if we find ourselves writing our own implementation of concepts like 'event with start and end time' and actions like 'return a view on all events with property X', then we could consider using an existing calendaring API like Google Calendar to do this for us. So the basic entity of a 'class' would exist in GCal and the ESP-specific relations between a class and a student, etc. would exist in our website. Additional ESP-specific properties for these GCal events could be specified with something like [private extended properties](https://developers.google.com/calendar/extended-properties).
> 
> I don't know if this is actually a good idea at all. Maybe the only thing it makes easier is the front end. woman_shrugging

I think this is all an interesting idea, though I think the main reason I wouldn't want this is that GCal's Event doesn't look exactly like what we want? It seems a bit complicated / it solves a different problem for Google than it solves for us. There seem to be a bunch of other properties that would just be leftover junk and if Google decides to change how things work, we could be in some trouble. That said, this is probably a good thing to keep in mind (maybe later when events are more fleshed out?)

server/core/models/clazz.py
```diff
@@ -48,6 +48,10 @@ def program(self):
     def num_students(self):
         return self.studentclassregs.count()
 
+    @property
+    def scheduledblocks(self):
+        return self.scheduledblock_set.all()
+
```
evantey14, line 54: Now that we have related_names, I think this isn't necessary?

server/core/views/studentreg.py
```diff
@@ -34,23 +34,27 @@ class StudentProgramClasses(APIView):
...
-                None,
-            ],
-            "timeslots": ["Sat 10-11", "Sat 11-12", "Sat 12-1"],
-        }
+        prog = Program.objects.get(name=program, edition=edition)
+        studentreg = StudentRegistration.objects.get(student=user, program=prog)
+        if "include_empty_timeslots" in params and params["include_empty_timeslots"]:
+            schedule = studentreg.get_schedule(include_empty_timeslots=True)
+        else:
+            schedule = studentreg.get_schedule()
```
evantey14, line 48: This is a great example of a place where `.get()` is useful. You should be able to just do `include_empty_timeslots = params.get("include_empty_timeslots", False)` then `schedule = studentreg.get_schedule(include_empty_timeslots=include_empty_timeslots)`

server/core/models/registration.py
```diff
@@ -39,19 +39,29 @@ class StudentRegistration(models.Model):
...
-        return Class.objects.filter(id__in=ids)
-        # TODO clean this type of query up with select_related
-
     class Meta:
         unique_together = (("student", "program"),)
 
+    @property
+    def sections(self):
+        sections = self.classregs.values_list("section", flat=True)
+        return Section.objects.filter(pk__in=sections)
```
evantey14, line 48: I feel like my intuition says there's a better way to do this than to invoke the Section manager? But we don't have to fix this now

server/core/models/clazz.py
mvadari: Done

server/core/views/studentreg.py
mvadari: Oh lit. Done.

server/core/models/registration.py
mvadari: I did a bunch of Googling for this when I was looking into PR #39 and couldn't find anything. I'll leave a TODO for now.

PR #23, opened 2020-07-12
Implement visual section selection in class catalog

mvadari:
Add a dropdown to each class in the catalog to select the relevant section.

Doesn't do anything functional yet (that's the next PR), just shows the dropdown, and sends a message to the console when a section is clicked on.

Test:
* Made sure the dropdown rendered for all classes
* Made sure the dropdown rendered the correct timeslot/text for each section
* Created a class with >1 section and made sure all the sections rendered correctly

image

### DIFF
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -16,7 +16,6 @@ type State = {
   displayOnlyOpenClasses: boolean;
   enrolledClasses: string[];
   timeslots: string[];
-  catalog: Clazz[];
 };
 
 // helper functions
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -69,7 +69,7 @@ class ClassChangesDashboard extends Component<Props, State> {
         });
       });
   }
-  
+
   renderClassSchedule() {
     return (
       <div className="column is-6">
@@ -119,13 +119,13 @@ class ClassChangesDashboard extends Component<Props, State> {
 
   addWaitlistSection(e: React.MouseEvent, clazz: Class) {
     e.preventDefault(); // TODO use button instead of anchor so this isn't needed
-    console.log("Adding to waitlist " + clazz.title);
+    console.log(`Adding to waitlist ${clazz.title}`);
   }
 
   removeSection(e: React.MouseEvent, section: Section) {
     e.preventDefault(); // TODO use button instead of anchor so this isn't needed
     if (section) {
-      console.log("Removing " + section.number);
+      console.log(`Removing ${section.number}`);
       // TODO make this functional
       // TODO refresh page or API calls
     }
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -2,6 +2,7 @@ import React, { Component } from "react";
 import axiosInstance from "../axiosAPI";
 import { Class, Section } from "./types";
 import { studentScheduleEndpoint, classCatalogEndpoint } from "../apiEndpoints";
+import "./Catalog.css";
 import { renderCustomInput } from "../forms/helpers";
 
 type Props = {
@@ -133,9 +134,9 @@ class ClassChangesDashboard extends Component<Props, State> {
   renderAddClassDropdown(clazz: Class) {
     return (
       <div className="card-footer-item">
-        <div className="dropdown is-hoverable mr-4">
-          <div className="dropdown-trigger">
-            <button className="button">
+        <div className="dropdown is-hoverable is-fullwidth">
+          <div className="dropdown-trigger is-fullwidth">
+            <button className="button is-fullwidth">
               <span>Sections</span>
               {/*TODO show selected section*/}
               <span className="icon is-small">
@@ -162,9 +163,6 @@ class ClassChangesDashboard extends Component<Props, State> {
         </div>
         {/*TODO replace button with "waitlist" button if no space in the chosen section*/}
         {/*TODO add relevant text */}
-        <button className="button" onClick={e => this.addSection(e, clazz)}>
-          Add class
-        </button>
       </div>
     );
   }
@@ -211,6 +209,9 @@ class ClassChangesDashboard extends Component<Props, State> {
               ? `${clazz.sections[0].num_students}/${clazz.capacity} students`
               : // TODO this shouldn't refer to just the first section
                 "Class is full"}
+            <button className="button ml-4" onClick={e => this.addSection(e, clazz)}>
+              Add class
+            </button>
           </h3>
         </div>
       </div>
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -16,6 +16,7 @@ type State = {
   displayOnlyOpenClasses: boolean;
   enrolledClasses: string[];
   timeslots: string[];
+  catalog: Clazz[];
 };
 
 // helper functions
@@ -66,6 +67,14 @@ class ClassChangesDashboard extends Component<Props, State> {
           catalog: res.data,
         });
       });
+  }
+
+  getClassCatalog() {
+    axiosInstance.get(`/${this.props.program}/${this.props.edition}/catalog/`).then(res => {
+      this.setState({
+        catalog: res.data,
+      });
+    });
   }
 
   renderClassSchedule() {
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -1,6 +1,6 @@
 import React, { Component } from "react";
 import axiosInstance from "../axiosAPI";
-import { Class } from "./types";
+import { Class, Section } from "./types";
 import { studentScheduleEndpoint, classCatalogEndpoint } from "../apiEndpoints";
 import { renderCustomInput } from "../forms/helpers";
 
@@ -59,6 +59,7 @@ class ClassChangesDashboard extends Component<Props, State> {
   }
 
   setupClassCatalog() {
+    // TODO do more detailed JSON parsing
     axiosInstance
       .get(`/${this.props.program}/${this.props.edition}/${classCatalogEndpoint}`)
       .then(res => {
@@ -108,16 +109,64 @@ class ClassChangesDashboard extends Component<Props, State> {
     // TODO will compare internal list of classes and catalog
   }
 
-  addClass(e: React.MouseEvent, clazz: Class) {
+  addSection(e: React.MouseEvent, clazz: Class) {
     e.preventDefault(); // TODO use button instead of anchor so this isn't needed
-    console.log("Adding " + clazz.title);
-    // TODO make this functional
+    console.log(`Adding ${clazz.title}`);
+    // TODO collect section number
+    // TODO refresh page or API calls
   }
 
-  removeClass(e: React.MouseEvent, clazz: Class) {
+  addWaitlistSection(e: React.MouseEvent, clazz: Class) {
     e.preventDefault(); // TODO use button instead of anchor so this isn't needed
-    console.log("Removing " + clazz.title);
-    // TODO make this functional
+    console.log("Adding to waitlist " + clazz.title);
+  }
+
+  removeSection(e: React.MouseEvent, section: Section) {
+    e.preventDefault(); // TODO use button instead of anchor so this isn't needed
+    if (section) {
+      console.log("Removing " + section.number);
+      // TODO make this functional
+      // TODO refresh page or API calls
+    }
+  }
+
+  renderAddClassDropdown(clazz: Class) {
+    return (
+      <div className="card-footer-item">
+        <div className="dropdown is-hoverable mr-4">
+          <div className="dropdown-trigger">
+            <button className="button">
+              <span>Sections</span>
+              {/*TODO show selected section*/}
+              <span className="icon is-small">
+                <i className="fas fa-angle-down"></i>
+              </span>
+            </button>
+          </div>
+          <div className="dropdown-menu">
+            <div className="dropdown-content">
+              {clazz.sections.map((section, index) => {
+                return (
+                  <a
+                    href="#void"
+                    key={index}
+                    className="dropdown-item"
+                    onClick={e => this.addSection(e, clazz)}
+                  >
+                    {section.scheduled_blocks.join(" / ")}
+                  </a>
+                );
+              })}
+            </div>
+          </div>
+        </div>
+        {/*TODO replace button with "waitlist" button if no space in the chosen section*/}
+        {/*TODO add relevant text */}
+        <button className="button" onClick={e => this.addSection(e, clazz)}>
+          Add class
+        </button>
+      </div>
+    );
   }
 
   toggleClassDescription(e: React.MouseEvent) {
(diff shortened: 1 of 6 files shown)


client/src/registration/types.ts
```diff
@@ -13,6 +13,7 @@ enum RegStatusOption {
 type Section = {
   num_students: number;
   number: number;
+  scheduledblock_set: string[];
```
wishdasher, line 16: `camelCase`

mvadari: Currently for JSON processing I'm doing everything as Javascript automatically does it (I'm thinking about how to best refactor so that JSON processing is more intentional). This means that these fields need to match exactly what the API shows, and Python uses `snake_case`. What's the tradeoff here? It has to be `snake_case` or `camelCase` in both places.

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -109,16 +111,61 @@ class StudentRegDashboard extends Component<Props, State> {
     // TODO will compare internal list of classes and catalog
   }
 
-  addClass(e: React.MouseEvent, clazz: Class) {
+  addSection(e: React.MouseEvent, clazz: Class) {
     e.preventDefault(); // TODO use button instead of anchor so this isn't needed
-    console.log("Adding " + clazz.title);
-    // TODO make this functional
+    console.log(`Adding ${clazz.title}`);
```
shardulc, line 116: Random thought not related to this PR: should we be using an actual logging system that is not just print to console? I.e. one where you can log stuff at different warning levels, output to a file(s), record where the log entry is coming from, etc. If this comment gets 2 or more upvotes I'll make a new issue for it.

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -109,16 +111,61 @@ class StudentRegDashboard extends Component<Props, State> {
...
+    // TODO collect section number
+    // TODO refresh page or API calls
   }
 
-  removeClass(e: React.MouseEvent, clazz: Class) {
+  addWaitlistSection(e: React.MouseEvent, clazz: Class) {
     e.preventDefault(); // TODO use button instead of anchor so this isn't needed
-    console.log("Removing " + clazz.title);
-    // TODO make this functional
+    console.log("Adding to waitlist " + clazz.title);
```
shardulc, line 123: reminder to use the `${clazz.title}` style consistently

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -109,16 +111,61 @@ class StudentRegDashboard extends Component<Props, State> {
...
+  removeSection(e: React.MouseEvent, section: Section) {
+    e.preventDefault(); // TODO use button instead of anchor so this isn't needed
+    if (section) {
+      console.log("Removing " + section.number);
+      // TODO make this functional
+      // TODO refresh page or API calls
+    }
+  }
+
+  renderAddClassDropdown(clazz: Class) {
```
shardulc, line 135: same comment as before, about having all the rendering stuff (HTML-generating functions) in a separate place

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -109,16 +111,61 @@ class StudentRegDashboard extends Component<Props, State> {
...
+    return (
+      <div className="card-footer-item">
+        <div className="dropdown is-hoverable is-fullwidth">
+          <div className="dropdown-trigger is-fullwidth">
+            <button className="button is-fullwidth">
+              <span>Sections</span>
+              {/*TODO show selected section*/}
+              <span className="icon is-small">
+                <i className="fas fa-angle-down"></i>
```
shardulc, line 144: italics... down arrow?

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -109,16 +111,61 @@ class StudentRegDashboard extends Component<Props, State> {
...
+                    href="#void"
+                    key={index}
+                    className="dropdown-item"
+                    onClick={e => this.addSection(e, clazz)}
+                  >
+                    {section.scheduled_blocks.join(" / ")}
+                  </a>
+                );
+              })}
+            </div>
```
shardulc, line 162: Hmm... so I'm getting more and more uncomfortable with this whole generate-HTML-in-JS thing. Isn't there a way to use template files for this? Like, you have a file called `template-div-addClassDropdown.html` that contains just this chunk of HTML, except that the parts that have to be filled in are `${placeholder}`. And in the JS code you just read in this file, replace the placeholder with the intended HTML (preferably using the equivalent of Python's `.format()`), and return it.

This way I can actually have a static, JS-free version of the entire website where all the placeholders are dummies. This is useful for trying out layout and UI changes without requiring the corresponding JS to be functional, and (if you have enough manpower) can parallelize development.

server/core/models/program.py
```diff
@@ -68,12 +68,9 @@ def __str__(self):
...
         return (
-            str(self.program)
-            + " ("
-            + self.start.strftime(date_format)
+            self.start.strftime(date_format)
             + ", "
             + self.start.strftime(time_format)
             + " to "
             + self.end.strftime(time_format)
-            + ")"
```
shardulc, line 78: oh dearie can we please use [Python's built-in string formatting](https://docs.python.org/3.5/library/string.html#formatspec) that even has special provisions for `strftime` times

client/src/registration/ChangeClassesDashboard.tsx
wishdasher: Yes, that's part of the dev plan (one of the things we're aware of), but I don't think it's quite worth tackling yet.

client/src/registration/ChangeClassesDashboard.tsx
wishdasher: Icon

client/src/registration/ChangeClassesDashboard.tsx
wishdasher: This is just how React and most JS libraries work. We can do some server-side rendering for the info pages, if you want to look into that.

> This is useful for trying out layout and UI changes without requiring the corresponding JS to be functional

Why would the corresponding JS not be functional?

client/src/registration/ChangeClassesDashboard.tsx
mvadari: Done

client/src/registration/ChangeClassesDashboard.tsx
mvadari: What comment was this?

server/core/models/program.py
mvadari: Done, but I think it makes it more difficult to understand what's happening.

client/src/registration/ChangeClassesDashboard.tsx
shardulc: about putting rendering functions in a separate file, like you did in #29

mvadari: I would agree, but this `render` function isn't a general function, it's a very specific one, and not likely to be used anywhere else.

client/src/registration/types.ts
evantey14: Looks like there are some libraries we might be able to use to convert between the two -- maybe this could be worked into the serializers in Django

mvadari: Added it to Trello.

server/core/models/program.py
```diff
@@ -65,15 +65,4 @@ class Timeslot(models.Model):
...
-            str(self.program)
-            + " ("
-            + self.start.strftime(date_format)
-            + ", "
-            + self.start.strftime(time_format)
-            + " to "
-            + self.end.strftime(time_format)
-            + ")"
-        )
+        return "{:%-m/%d/%y, %-I:%M %p} to {:%-I:%M %p}".format(self.start, self.end)
```
evantey14, line 68: Let's leave a docstring giving an example of what this looks like decoded?

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -156,25 +203,15 @@ class ClassChangesDashboard extends Component<Props, State> {
...
-          )}
+          {this.renderAddClassDropdown(clazz)}
           <h3 className="card-footer-item">
             {classHasSpace
               ? `${clazz.sections[0].num_students}/${clazz.capacity} students`
               : // TODO this shouldn't refer to just the first section
                 "Class is full"}
+            <button className="button ml-4" onClick={e => this.addSection(e, clazz)}>
+              Add class
+            </button>
```
evantey14, line 214: I'm not sure I understand why this is here if people are supposed to click on the sections to add them?

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -108,16 +110,61 @@ class ClassChangesDashboard extends Component<Props, State> {
...
+    // TODO refresh page or API calls
   }
 
-  removeClass(e: React.MouseEvent, clazz: Class) {
+  addWaitlistSection(e: React.MouseEvent, clazz: Class) {
     e.preventDefault(); // TODO use button instead of anchor so this isn't needed
-    console.log("Removing " + clazz.title);
-    // TODO make this functional
+    console.log(`Adding to waitlist ${clazz.title}`);
+  }
```
evantey14, line 123: Can we leave off this addition until an actual waitlist PR?

evantey14 (commented):
So I'm going to defer to @wishdasher for approval since I don't know how frontend dev process works, but personally I feel like my intuition says we shouldn't add unnecessary visual things unless they have a functional purpose. 

Is the goal of this PR just to display all the sections for a class? If so, I feel like we don't need dropdown CSS (unless we consider the dropdown interaction a core feature of our change classes design?). Also, if that's the goal, I'd prefer not having the changes in the various add/remove functions cause those aren't immediately related to the purpose of this PR. 

I think there are enough weird things with this ui (why are the dropdowns buttons / why do they dropdown when you hover? why is there an Add class button? why are the capacity numbers at the class level and not the section level?) (I forget how much of this was there before...) but yeah I feel like I would rather have a more concrete picture of what we want this to end up looking like / what steps we should take to get there than making this change and leaving all those weird loose ends.

wishdasher:
> So I'm going to defer to @wishdasher for approval since I don't know how frontend dev process works, but personally I feel like my intuition says we shouldn't add unnecessary visual things unless they have a functional purpose.
> 
> Is the goal of this PR just to display all the sections for a class? If so, I feel like we don't need dropdown CSS (unless we consider the dropdown interaction a core feature of our change classes design?). Also, if that's the goal, I'd prefer not having the changes in the various add/remove functions cause those aren't immediately related to the purpose of this PR.
> 
> I think there are enough weird things with this ui (why are the dropdowns buttons / why do they dropdown when you hover? why is there an Add class button? why are the capacity numbers at the class level and not the section level?) (I forget how much of this was there before...) but yeah I feel like I would rather have a more concrete picture of what we want this to end up looking like / what steps we should take to get there than making this change and leaving all those weird loose ends.

Agreed. I haven't been keeping close enough track of what was going on with all this, but it seems more productive to design UI flows and requesting review and feedback on those rather than forging ahead adding features in ways that may prove to be more rigid than is nice when we have to change these flows and interactions.

Adding mostly independent components and features without tying them too strongly with existing code feels better, for instance.

mvadari:
> > So I'm going to defer to @wishdasher for approval since I don't know how frontend dev process works, but personally I feel like my intuition says we shouldn't add unnecessary visual things unless they have a functional purpose.
> > Is the goal of this PR just to display all the sections for a class? If so, I feel like we don't need dropdown CSS (unless we consider the dropdown interaction a core feature of our change classes design?). Also, if that's the goal, I'd prefer not having the changes in the various add/remove functions cause those aren't immediately related to the purpose of this PR.
> > I think there are enough weird things with this ui (why are the dropdowns buttons / why do they dropdown when you hover? why is there an Add class button? why are the capacity numbers at the class level and not the section level?) (I forget how much of this was there before...) but yeah I feel like I would rather have a more concrete picture of what we want this to end up looking like / what steps we should take to get there than making this change and leaving all those weird loose ends.
> 
> Agreed. I haven't been keeping close enough track of what was going on with all this, but it seems more productive to design UI flows and request review and feedback on those rather than forging ahead adding features in ways that may prove to be more rigid than is nice when we have to change these flows and interactions.
> 
> Adding mostly independent components and features without tying them too strongly with existing code feels better, for instance.
> 
> Going back to this, should really push espider to give us a general spec.

I think I might have actually erred too much on the side of making small PRs, and also I should have done a better job of explaining where this was going. My overall objective here was to make a fully functional class catalog "module", with section selection and adding sections and stuff - basically everything that's necessary for a class catalog to fully function. 

The next draft PR after this uses the sections dropdown/what section has been selected to change the capacity/number of students information in the card (and changing the "add class" button to an "add waitlist" button if the class is full), and if no section is selected, then it just shows the class capacity. I think I might have made things a little confusing by calling `addSection` on the dropdown itself. I can combine those two PRs if you think that would make more sense, even if it means more code to review.

server/core/models/program.py
mvadari: Done.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: See comment below. I changed what clicking on the section does (just prints out what is the chosen section) to make it more clear.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: (y)

evantey14:
> > > So I'm going to defer to @wishdasher for approval since I don't know how frontend dev process works, but personally I feel like my intuition says we shouldn't add unnecessary visual things unless they have a functional purpose.
> > > Is the goal of this PR just to display all the sections for a class? If so, I feel like we don't need dropdown CSS (unless we consider the dropdown interaction a core feature of our change classes design?). Also, if that's the goal, I'd prefer not having the changes in the various add/remove functions cause those aren't immediately related to the purpose of this PR.
> > > I think there are enough weird things with this ui (why are the dropdowns buttons / why do they dropdown when you hover? why is there an Add class button? why are the capacity numbers at the class level and not the section level?) (I forget how much of this was there before...) but yeah I feel like I would rather have a more concrete picture of what we want this to end up looking like / what steps we should take to get there than making this change and leaving all those weird loose ends.
> > 
> > 
> > Agreed. I haven't been keeping close enough track of what was going on with all this, but it seems more productive to design UI flows and request review and feedback on those rather than forging ahead adding features in ways that may prove to be more rigid than is nice when we have to change these flows and interactions.
> > Adding mostly independent components and features without tying them too strongly with existing code feels better, for instance.
> > Going back to this, should really push espider to give us a general spec.
> 
> I think I might have actually erred too much on the side of making small PRs, and also I should have done a better job of explaining where this was going. My overall objective here was to make a fully functional class catalog "module", with section selection and adding sections and stuff - basically everything that's necessary for a class catalog to fully function.
> 
> The next draft PR after this uses the sections dropdown/what section has been selected to change the capacity/number of students information in the card (and changing the "add class" button to an "add waitlist" button if the class is full), and if no section is selected, then it just shows the class capacity. I think I might have made things a little confusing by calling `addSection` on the dropdown itself. I can combine those two PRs if you think that would make more sense, even if it means more code to review.

Ah okay yeah I think at least having the addSection functionality work here would be nice. 

I still think addWaitlist isn't something we should be too worried about yet (as long as we have it written down on a card somewhere for the future, but I'm not sure if it should be an MVP feature). 

Then last (just my thoughts so feel free to take it or leave it) I personally am not a huge fan of the dropdown display. Again, idk how people normally do frontend, but I feel like we should focus on just displaying the correct information first and then think later about the subtleties of how the info gets displayed later. I think particularly for displaying section, there doesn't seem to be a huge difference between just displaying a list of elements and using dropdown menus. Since odds are that we're probably gonna change how this works, I would opt for the code that's cleaner/simpler so when we edit later it's more straightforward.

mvadari:
Should I switch to showing sections in the catalog list instead of classes, for now, before we figure out the details of the class changes UX experience?

PR #27, opened 2020-07-12
Properly hash passwords when inputting sample data

mvadari:
Name says it all. 

Also added the root account, since we all create it anyways, and passwords are being hashed properly. 

TEST:
* Examined the passwords in the shell and confirmed that they looked hashed
* Logged into the admin panel with the `edward` account (which previously didn't work)

### DIFF
--- a/server/sample_data/ESPUser.tsv
+++ b/server/sample_data/ESPUser.tsv
@@ -11,4 +11,4 @@ id	username	password	is_staff	is_superuser
 1023	teacher3	password	False	False
 1024	teacher4	password	False	False
 1040	edward	lobster	True	True
-1	root	root	True	True
+999	root	root	True	True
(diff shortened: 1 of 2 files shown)


server/sample_data/ESPUser.tsv
```diff
@@ -11,3 +11,4 @@ id	username	password	is_staff	is_superuser
 1023	teacher3	password	False	False
 1024	teacher4	password	False	False
 1040	edward	lobster	True	True
+1	root	root	True	True
```
evantey14, line 14: not sure if we want to pick a higher id to avoid id ocnflicts with any existing db users?

mvadari: Yeah fair, I figured we'd use this to do the initial population so I wasn't worried about it, but I'll change it just to be safe.

PR #28, opened 2020-07-13
Fix admin panel favicon error

mvadari:
Every time I opened up the admin panel, there would be a 500 error on the favicon, because it couldn't find it in the directory. This is a fix for that error. There is now favicon generation for the admin panel.


wishdasher:
I'm merging this. Next time deployment happens, let's check to make sure it works as expected.

wishdasher:
I'm merging this. Next time deployment happens, let's check to make sure it works as expected.

PR #29, opened 2020-07-13
Implement searching for classes

mvadari:
Searches title and description for the search string. 

Added message to display if there are no classes that fit the search string. 

Test: searched for a bunch of different strings, made sure the results were as expected. 

image
image


server/core/views/api.py
```diff
@@ -70,8 +71,16 @@ class ClassCatalog(APIView):
     """
 
     def get(self, request, program, edition, format=None):
+        params = request.GET
+
         prog = Program.objects.get(name=program, edition=edition)
         classes = Class.objects.filter(program=prog)
+
+        if "search" in params:
+            query = params["search"]
```
evantey14, line 80: Another way to do all of this would've been `query = request.GET.get("search")` then testing `if query is not None`.

server/core/views/api.py
```diff
@@ -70,8 +71,16 @@ class ClassCatalog(APIView):
...
+            classes = classes.filter(Q(title__icontains=query) | Q(description__icontains=query))
+            # TODO investigate speed of query, consider postgres.search
```
evantey14, line 82: I personally don't think we need this TODO (not that it hurts anyone if we leave it here), but if we were to eventually care about speed, we should probably be profiling lots of different code to see figure out what should be optimized anyway (I'm not sure if this TODO adds much on top of what would be found by the profiling).

server/core/views/api.py
```diff
@@ -70,8 +71,16 @@ class ClassCatalog(APIView):
...
 
     def get(self, request, program, edition, format=None):
+        params = request.GET
+
         prog = Program.objects.get(name=program, edition=edition)
         classes = Class.objects.filter(program=prog)
+
+        if "search" in params:
+            query = params["search"]
+            classes = classes.filter(Q(title__icontains=query) | Q(description__icontains=query))
```
evantey14, line 81: This is probably okay (I think it's mostly readable), but you could consider trying to put this in a well-named Manager method

mvadari:
> From a higher-level design standpoint, by doing this, we're moving towards the decision of more smaller API requests for subsets of classes. This means our backend will be doing more work in terms of sending back results for any request the frontend makes. I think this is only really worth it if searching on the frontend is slow/expensive for clients. I'm not really sure how expensive that is, so I'm not saying we should change, but this tradeoff is probably a good thing to keep in mind (we shouldn't write code that corners us into this approach too much).

To be honest, I didn't do much analysis on this. But it seems much easier/faster to do searches like this on the backend, using SQL, than by trying to filter with Javascript. Maybe @wishdasher has more insight on this.

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -60,8 +66,18 @@ class StudentRegDashboard extends Component<Props, State> {
   }
 
   setupClassCatalog() {
+    this.sendClassCatalogRequest();
+  }
+
+  sendClassCatalogRequest(includeSearch: boolean = false) {
```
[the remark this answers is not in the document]
mvadari: Moved all the logic from `sendClassCatalogRequest` into `setupClassCatalog` - does that work?

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -60,8 +66,18 @@ class StudentRegDashboard extends Component<Props, State> {
...
+    const params = {} as any;
```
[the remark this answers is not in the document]
mvadari: Yeah we could, I was lazy. Fixed.

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -60,8 +66,18 @@ class StudentRegDashboard extends Component<Props, State> {
...
+    if (includeSearch) {
+      params["search"] = this.state.classSearchQuery;
```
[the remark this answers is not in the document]
mvadari: Sure, I'll change that.

server/core/views/api.py
mvadari: Should I change to that?

server/core/views/api.py
mvadari: I put this TODO in because while I was doing research about how to search the db with django, a lot of people said that using `icontains` can be really inefficient because it uses `LIKE` in SQL, and recommended using `postgres.search`.

server/core/views/api.py
mvadari: Added a TODO

wishdasher:
> > From a higher-level design standpoint, by doing this, we're moving towards the decision of more smaller API requests for subsets of classes. This means our backend will be doing more work in terms of sending back results for any request the frontend makes. I think this is only really worth it if searching on the frontend is slow/expensive for clients. I'm not really sure how expensive that is, so I'm not saying we should change, but this tradeoff is probably a good thing to keep in mind (we shouldn't write code that corners us into this approach too much).
> 
> To be honest, I didn't do much analysis on this. But it seems much easier/faster to do searches like this on the backend, using SQL, than by trying to filter with Javascript. Maybe @wishdasher has more insight on this.

This is querying through class titles and descriptions? We have, what, around 500 classes for a Splash, fewer for every other program? Classes should be fairly set once students are looking at them, with the exception of available space. I say we should return all relevant class data (for now: title and description) given a program-variant pair, cache the request (if we ever get to caching) and filter on the frontend. Availability could be handled in a separate call.

I think most of the catalog will be displayed to a given student anyways. The queries will cover a majority of the relevant data.

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -31,6 +35,12 @@ function renderLinkedText(displayedText: string, url: string) {
   );
 }
 
+function renderTextInSection(displayedText: string, centered = false) {
+  return (
+    <h3 className={"is-size-5" + (centered ? " has has-text-centered" : "")}>{displayedText}</h3>
```
evantey14, line 40: Is `has` a real class? (I googled for like 30 seconds and couldn't figure it out, so here we are.

server/core/views/api.py
evantey14: Up to you? I think it's okay as is. If I were writing, this^ is probably what I would've done, but you should feel free to judge what you think looks cleaner.

mvadari:
Changed processing of the search to be in the frontend, instead of via API query.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: Oops, typo

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -173,8 +180,14 @@ class StudentRegDashboard extends Component<Props, State> {
...
 
   submitSearch = (e: React.MouseEvent<HTMLButtonElement>) => {
-    console.log("Searching for " + this.state.classSearchQuery);
-    // TODO make this functional
+    console.log(`submit search for ${this.state.classSearchQuery}`);
+    this.setState({
+      displayedCatalog: this.state.catalog.filter(
+        clazz =>
+          clazz.title.toLowerCase().indexOf(this.state.classSearchQuery) >= 0 ||
+          clazz.description.toLowerCase().indexOf(this.state.classSearchQuery) >= 0
```
shardulc, line 188: Shouldn't the search term also be lowercased? Should it be stripped of whitespace, non-alphanumeric characters, and should searching for 'singing' turn up 'Sing with the Toons'? Should searching for 'web' turn up 'An Introduction to Weber's Law'?

Also, I expect that in the future there will be an option to search by title only or description only or more parameters. Might be nice to build that into the function spec to begin with.

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -215,8 +228,9 @@ class StudentRegDashboard extends Component<Props, State> {
           Only open classes
         </button>
         {/*TODO add more filters*/}
-
-        {this.state.catalog.map(clazz => this.renderClass(clazz))}
+        {this.state.catalog.length > 0
+          ? this.state.displayedCatalog.map(clazz => this.renderClass(clazz))
```
shardulc, line 232: does `map(this.renderClass)` not work

shardulc: or `map(this.renderClass, this.state.displayedCatalog)` like a good nice functional language

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -215,8 +228,9 @@ class StudentRegDashboard extends Component<Props, State> {
...
+          : renderTextInSection("No classes available.", true)}
```
shardulc, line 233: "No classes match your search term."

server/core/views/api.py
```diff
@@ -72,6 +72,7 @@ class ClassCatalog(APIView):
     def get(self, request, program, edition, format=None):
         prog = Program.objects.get(name=program, edition=edition)
         classes = Class.objects.filter(program=prog)
+
```
shardulc, line 75: seems weird, is this the style you are following

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -31,12 +32,17 @@ function renderLinkedText(displayedText: string, url: string) {
   );
 }
 
+function renderTextInSection(displayedText: string, centered = false) {
```
shardulc, line 35: This function is used exactly one. Should it be inlined? Also, maybe a broader-ranging code organization opinion: you might want to put all the 'rendering' code in a separate file, like all the functions that actually have hardcoded HTML in them, because then you can have a central place to make changes like "oh actually all the things we're currently using `<h3>` for should be `<spans>` with this custom CSS class"

client/src/registration/ChangeClassesDashboard.tsx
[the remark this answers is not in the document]
mvadari: That change is in the pipeline, and hasn't been implemented yet.

mvadari: Yeah, it should be lowercased - I missed that, oops. 

I'll add a TODO for the other things you mentioned, to make a better search experience - I think that's a little too complicated to be worth it for now.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: The former does not work. I assume the latter will not either.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: I was trying to come up with some phrase that is neutral, that could represent either no classes at all (for some reason, maybe if someone accesses the website too early or something) or no classes by the search term/other filter parameters. Maybe that's too broad and I should separate out "no classes" in total and "no classes by these filter parameters". Thoughts?

server/core/views/api.py
mvadari: It's neater post-refactor-that-just-got-pushed. I just pushed this.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: Fair, put it all in a separate file.

client/src/registration/ChangeClassesDashboard.tsx
wishdasher: It should work if you bind the method.
`renderClass = (clazz: Class) => `

client/src/registration/ChangeClassesDashboard.tsx
wishdasher: > that could represent either no classes at all
This shouldn't be the place where they discover that, so I think Shardul's suggestion is fine.

client/src/registration/ChangeClassesDashboard.tsx
wishdasher: Oh wait, I may be thinking about passing functions as props.

wishdasher: Try it anyways?

mvadari: Ok that worked - done.

client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -189,8 +181,15 @@ class StudentRegDashboard extends Component<Props, State> {
...
 
   submitSearch = (e: React.MouseEvent<HTMLButtonElement>) => {
-    console.log("Searching for " + this.state.classSearchQuery);
-    // TODO make this functional
+    console.log(`submit search for ${this.state.classSearchQuery}`);
+    this.setState({
+      displayedCatalog: this.state.catalog.filter(
+        clazz =>
+          clazz.title.toLowerCase().indexOf(this.state.classSearchQuery.toLowerCase()) >= 0 ||
+          clazz.description.toLowerCase().indexOf(this.state.classSearchQuery.toLowerCase()) >= 0
```
evantey14, line 189: Maybe consider defining functions somewhere like `classContainsText(class, query)` that does this check for readability purposes

client/src/registration/ChangeClassesDashboard.tsx
evantey14: I think I mildly agree with Sophie / Shardul, but I don't think it's a huge deal and am okay with either.

mvadari: (y) - changed to `search filters` so it would be more applicable for later additional filters.

server/core/views/studentreg.py
```diff
@@ -8,7 +8,6 @@
 
 # Registration API calls
 
-
```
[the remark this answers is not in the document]
mvadari: Was just too much unnecessary space in the file.

client/src/registration/ChangeClassesDashboard.tsx
mvadari: Done.

client/src/helperTextFunctions.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+function renderTextInSection(displayedText: string, centered = false) {
+  return (
+    <h3 className={"is-size-5" + (centered ? " has has-text-centered" : "")}>{displayedText}</h3>
```
evantey14, line 5: nit: I forget what happened last time, is the first `has` a typo?

mvadari:
> As far as next steps, this file has gotten a bit bloated/messy. Hopefully refactoring towards using react hooks & maybe making a Class component will help organize the logic better.

Yeah, once this set of PRs has been pushed through I was gonna refactor and make StudentSchedule and ClassCatalog components. But dealing with conflicts would be a pain if I did it before we finished the PRs.

client/src/helperTextFunctions.tsx
mvadari: Yeah, that was probably in another PR but I'll handle it here too.

PR #30, opened 2020-07-13
Reorganize API file into multiple separate files

mvadari:
No new code, just reorganizing old code so it's easier to navigate and there isn't just one giant file for all the API code. 

**init:**
* class catalog (there's not enough other stuff related to this to make it its own category)
* star imports

**accounts:**
* login
* signup

**dashboard:**
* student dashboard
* teacher dashboard

**studentreg:**
* studentreg object
* student schedule
* update profile
* fill out all the forms

Test: derped around and made sure all the API calls still worked.


mvadari:
> Are we anticipating having other views? If not, I'd feel comfortable just putting this all in an api module under core or just directly in views. Not sure if there's a big reason to have a views module that only has api in it

Yeah fair, it just felt weird because of how Django is intended to be organized.

mvadari:
> When I go to /dashboard, get an error that get_previous_student programs doesn't exist?

Fixed, was because this and model organization happened at the same time and weren't properly reconciled.

PR #31, opened 2020-07-13
Related_names in Django models

mvadari:
Added defined `related_name`s and `related_query_name`s for all objects with `ForeignKey` relations

Restructured some functions to use the `related_name` structures (which cleaned it up)

Added a couple TODOs

Done based on comments in PR #21 

Test:
* Any function changed that is used in an API call was tested by running the UI and making sure that rendered properly
* Other changes were tested in the CLI - calling them, making sure they returned as they were supposed to.


server/core/models/classreg.py
```diff
@@ -5,8 +5,18 @@
...
+        on_delete=models.CASCADE,
+        related_name="classregs",
+        related_query_name="classreg",
+    )
+    section = models.ForeignKey(
+        Section,
+        on_delete=models.CASCADE,
+        related_name="studentclassregs",
+        related_query_name="studentclassreg",
+    )
```
evantey14, line 19: :/ I am not a fan of "classregs" and "studentclassregs". I can't think of anything better, so instead I'm just voicing my feelings. Maybe @EmilyCaragay can think of better words? Might be a good place to start code reviewing.

server/core/models/program.py
```diff
@@ -24,7 +24,7 @@ def __str__(self):
 
     @staticmethod
     def get_open_student_programs():
-        return Program.objects.all().filter(student_reg_open=True)
+        return Program.objects.filter(student_reg_open=True)
```
evantey14, line 27: The fact that we have "student_reg_open" with an underscore between student and reg is another reason I am not a fan of this mythical "studentreg" object. I guess it makes some sense cause this student registration is talking about being in a state where people are allowed to make StudentRegistration objects, while generally "studentreg" means the actual registration details associated with student who is registering for a program. But how can anyone know to use `studentreg` or `student_reg` correctly? (Again, not a concrete thing that needs to be fixed in this PR, but...yeah)

server/core/models/registration.py
```diff
@@ -81,9 +96,7 @@ class Meta:
 
     @property
     def classes(self):
-        ids = esp_models.TeacherClassRegistration.objects.filter(teacher=self).values_list(
-            "clazz", flat=True
-        )
+        ids = self.classregs.all().values_list("clazz", flat=True)
```
evantey14, line 99: Does this `all()` need to be here?

server/core/models/classreg.py
mvadari: Happy to take suggestions, couldn't think of better names myself.

server/core/models/registration.py
mvadari: Oops nope, took it out.

server/core/models/program.py
mvadari: I'll put discussing these things on the agenda for the next meeting. Maybe something can come out of that.

mvadari:
> Add a description about how you tested this?

Done.

PR #32, opened 2020-07-13
Refactor auth to use context and hooks

wishdasher:
I refactored authentication to use [React context](https://reactjs.org/docs/context.html). This means we no longer need to pass `loggedIn` as a prop everywhere, hooray. Currently, I have replaced it with an `AuthContext` represented by the access token. Because I implemented this with [React hooks](https://reactjs.org/docs/context.html), I had to refactor some of the affected components to use hooks instead of classes. They are functionally equivalent. Going forward, we should look into using hooks more.

Also added some TODOs involving error handling and refactoring. The login and logout logic should be independent of the UI components that trigger them, but I have left this alone for now.
 
**Testing**
- sign in with existing account
- create new user account
- try to log in with incorrect credentials

**Coming up next:**
- Remove `loggedIn` as a prop on all components that use it
- I encountered a render on mount bug earlier but can't reproduce it. If it crops up again, we will address it then.
- Protect private routes behind the `AuthContext`
- Redirect correctly when user tried to access a protected route without being logged in. Some of the groundwork for that is in this PR because it was useful to keep that design in mind when refactoring components.
- Extrapolate context pattern to get rid of passing around `username`, `isStudent`, and `isTeacher`


wishdasher:
Please test this more (with console open) to make sure I didn't break authentication routes and whatnot.

client/src/App.tsx
```diff
@@ -43,135 +45,116 @@ const NotFound = () => (
   </section>
 );
 
+// TODO: rename, see if we can just get rid of it
 type UserState = {
-  loggedIn: boolean;
   username: string;
   isStudent: boolean;
   isTeacher: boolean;
```
mvadari, line 52: Shouldn't this information also be in the context?

client/src/App.tsx
```diff
@@ -43,135 +45,116 @@ const NotFound = () => (
...
-  constructor(props) {
-    super(props);
-    this.state = {
-      loggedIn: localStorage.getItem("token") ? true : false,
-      username: "",
-      isStudent: false,
-      isTeacher: false,
-    };
-  }
+function App(props: {}) {
```
mvadari, line 55: Why switch to a function from a class?

client/src/App.tsx
```diff
@@ -43,135 +45,116 @@ const NotFound = () => (
...
+          {/* <PrivateRoute as={Nextup} path="private" /> */}
+          <LoginPage path="login" username={userInfo.username} setToken={setToken} />
+          <SignupPage path="signup" username={userInfo.username} setToken={setToken} />
+
+          <Dashboard
+            path="dashboard"
+            loggedIn={authToken}
+            username={userInfo.username}
+            isStudent={userInfo.isStudent}
+            isTeacher={userInfo.isTeacher}
```
mvadari, line 101: I thought the point of using contexts and hooks was to remove the need for this?

client/src/accounts/LoginForm.tsx
```diff
@@ -1,77 +1,81 @@
...
+          // test referrer in the future
+          navigate(referer);
+        } else {
+          console.log(`bad status ${result.status}`);
+          // TODO error state
+        }
+      })
+      .catch(e => {
+        // TODO error state
+        console.log(`login post error ${e}`);
```
mvadari, line 56: This should be in a separate PR (e.g. better error handling in API calls).

client/src/accounts/StudentSignupForm.tsx
```diff
@@ -58,9 +59,10 @@ class StudentSignupForm extends Component<Props, State> {
       .then(result => {
         // TODO: handle failure
         axiosInstance.defaults.headers["Authorization"] = "JWT " + result.data.tokens.access;
+        // TODO: add this back, and REFACTOR. This should not have been in two places.
         localStorage.setItem("token", result.data.tokens.access);
         localStorage.setItem("refresh", result.data.tokens.refresh);
-        this.props.onLogin({ loggedIn: true });
+        this.props.setToken(result.data.tokens.access);
```
mvadari, line 65: It looks like you're not saving the refresh token at all.  That's necessary for general auth - without it, you'll effectively only be logged in for 5 minutes.

client/src/App.tsx
mvadari: I think this should have been done in a separate PR.

client/src/App.tsx
```diff
@@ -43,135 +45,116 @@ const NotFound = () => (
...
-        this.setState({
+        setUserInfo({
           username: result.data.username,
           isStudent: result.data.is_student,
           isTeacher: result.data.is_teacher,
         });
       });
     }
-  }
+  }, [authToken]);
```
mvadari, line 76: What does this do?

client/src/App.tsx
wishdasher: No, I left that for the future. I mentioned this in https://github.com/mitesp/espider/pull/32#issue-448550865

client/src/App.tsx
wishdasher: The context design uses hooks, which are incompatible with classes.

client/src/App.tsx
wishdasher: See above.

client/src/accounts/LoginForm.tsx
wishdasher: This is something I noticed while refactoring for this PR. I didn't make any real changes here, just added the TODO. If you want me to create a separate PR to add in two TODO statements, I guess so but I'm against it.

client/src/accounts/StudentSignupForm.tsx
wishdasher: The token isn't being used for anything except to verify that the user is logged in. It's functioning as a boolean. The API calls still pull from local storage. Using the token instead of a boolean here gives us future flexibility if we decide to refactor API calls.

client/src/App.tsx
wishdasher: https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect

wishdasher:
@mvadari Could I get another look at this? (responses to comments)
If you want a pr with just class -> function conversions and another one for the context, I suppose that's possible.

client/src/App.tsx
wishdasher: Right, from next steps in the PR comment:
> Remove loggedIn as a prop on all components that use it

client/src/App.tsx
```diff
@@ -43,135 +45,116 @@ const NotFound = () => (
...
+  // where they currently live
+
+  return (
+    <AuthContext.Provider value={authToken}>
+      <Nav loggedIn={authToken} username={userInfo.username} setToken={setToken} />
+      <main className="px-3 py-5">
+        <Router>
+          <Home path="/" />
+          <Home path="logout" />
+          {/* <PrivateRoute as={Nextup} path="private" /> */}
```
evantey14, line 92: Seems like this is doing some protection under the authcontext? I'm fine with leaving it here if it's cleaned up soon, but would prefer either a TODO or just leaving it out until the PR where it's introduced.

client/src/accounts/LoginForm.tsx
```diff
@@ -1,77 +1,81 @@
-import React, { Component } from "react";
+import React, { useState } from "react";
 import axiosInstance from "../axiosAPI";
 import { renderStandardFormField } from "../forms/helpers";
 import { loginEndpoint } from "../apiEndpoints";
+
```
evantey14, line 5: nit: was this intentional?

client/src/accounts/LoginForm.tsx
```diff
@@ -1,77 +1,81 @@
...
-  onLogin: Function;
-};
+import { useAuth } from "../context/auth";
 
-type LoginState = {
-  username: string;
-  password: string;
+type Props = {
+  setToken: (arg0: string) => void;
+  location?: { state: { referer: string } };
```
evantey14, line 12: whoaa, TIL: https://en.wikipedia.org/wiki/HTTP_referer

client/src/accounts/LoginForm.tsx
```diff
@@ -1,77 +1,81 @@
...
-  constructor(props: LoginProps) {
-    super(props);
-    this.state = {
-      username: "",
-      password: "",
-    };
-  }
+  const authToken = useAuth();
+
+  // If there was no referred, default to the dashboard.
```
evantey14, line 21: nit: referred -> referrer?

client/src/accounts/LoginForm.tsx
evantey14: Also, is there a reason this is structured in this way? (as oppose to idk just `referer?: string`?)

client/src/accounts/LoginForm.tsx
```diff
@@ -1,77 +1,81 @@
...
-type LoginProps = {
-  onLogin: Function;
-};
+import { useAuth } from "../context/auth";
 
-type LoginState = {
-  username: string;
-  password: string;
+type Props = {
+  setToken: (arg0: string) => void;
```
evantey14, line 11: nit: are we able to make this just `token`?

client/src/accounts/LoginForm.tsx
```diff
@@ -1,77 +1,81 @@
...
-      <form onSubmit={this.handleLogin}>
-        <h1 className="has-text-centered is-size-3">Log in</h1>
-        {renderStandardFormField("username", this.handleChange, this.state.username)}
+  }
 
-        {renderStandardFormField("password", this.handleChange, this.state.password)}
+  // TODO: add in this redirect behavior, test with referrer later
+  if (authToken) {
+    navigate(referer);
+  }
```
evantey14, line 63: I feel like this maybe belongs in `LoginPage`?

client/src/accounts/StudentSignupForm.tsx
```diff
@@ -69,11 +71,8 @@ class StudentSignupForm extends Component<Props, State> {
     return (
       <form onSubmit={this.handleSignup}>
         {renderStandardFormField("username", this.handleChange, this.state.username)}
-
```
evantey14, line 72: nit: I'd prefer lint fixes like this to be in PRs focused on fixing lint, but idk eh

client/src/accounts/LoginForm.tsx
wishdasher: We should get import organization for js.

client/src/accounts/LoginForm.tsx
wishdasher: That's how the router package will pass it, come then.

client/src/accounts/LoginForm.tsx
wishdasher: Will deal with when I actually figure out referer.

PR #33, opened 2020-07-16
Add other global auth-related fields to AuthContext

wishdasher:
Testing: log in, sign up, etc.

Next up: removing those props from all components

### DIFF
--- a/client/src/context/auth.tsx
+++ b/client/src/context/auth.tsx
@@ -1,7 +1,18 @@
 import { createContext, useContext } from "react";
 
-export const AuthContext = createContext("");
+export const AuthContext = createContext({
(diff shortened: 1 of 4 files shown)


client/src/context/auth.tsx
```diff
@@ -1,7 +1,18 @@
...
+  isTeacher: false,
+});
 
 export function useAuth() {
   return useContext(AuthContext);
 }
+
+export function useLoggedIn() {
+  const auth = useContext(AuthContext);
+  return auth.token;
```
evantey14, line 17: Can we make this actually return a boolean? (if nothing else, I imagine typescript will eventually be mad?)

wishdasher: Oops yeah.

PR #35, opened 2020-07-16
Initialize scheduler app

mvadari:
Spin up an initial scheduler app, via Docker and React. Loads at `localhost:5000`

Doesn't display anything right now, just a shell.

Haven't tested that everything works with Heroku (kind of assuming it won't), will deal with that after. 

image


wishdasher:
Seems to be a lot of unnecessary code here? The nav bar, etc.

mvadari:
> Seems to be a lot of unnecessary code here? The nav bar, etc.

I was copying from the `client/` code. I kept the login code because we want this app to only be accessible by admins, so admins need to log in somehow, and therefore the nav bar made sense to keep in some form, so that there could be a login button (and presumably there will be other pages, like a list of classes or something).

evantey14:
Yeah so as per discord discussion, I'm gonna close this. You're free to just work in a branch off master. I imagine the work will be pretty independent of master since it'll probably focus on getting channels working after one large api dump.

PR #37, opened 2020-07-17
Make default reg status for StudentRegistration equal to its program's

mvadari:
Title says it all

Test:
* created object in admin panel, checked to make sure reg status equaled the program's
* created object in shell, checked to make sure reg status equaled the program's
* edited `reg_status` in the admin panel, made sure it changed properly (instead of refusing to change)

### DIFF
--- a/server/core/admin/__init__.py
+++ b/server/core/admin/__init__.py
@@ -146,7 +146,7 @@ class StudentRegistrationAdmin(admin.ModelAdmin):
 
     def add_view(self, request, extra_context=None):
         extra_context = extra_context or {}
-        self.fields = ("student", "program", "reg_status")
+        self.fields = ("student", "program")
         self.readonly_fields = ()
         self.inlines = []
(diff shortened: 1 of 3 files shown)


evantey14 (commented):
So, I guess first, I personally don't think the save function is the right place for this code. I think the better way to implement this is to override the `get_or_create` function in the StudentRegistration manager. Especially as our models get more complicated, I think we'll want to keep functionality outside of the save() method or else we'll just end up with a giant save method that does lots of validation / cleaning etc. (iirc this is a real problem lots of Django codebases have and Managers are the correct solution)

Secondly, at a higher level, I'm not sure if this is a change we want? Now that we have more people, maybe we can revisit the current plan for student registration flow. I think @wishdasher wasn't a fan of the independent FSM model, so I think we should talk it out before developing it out more.

mvadari:
> Now that we have more people, maybe we can revisit the current plan for student registration flow. 

Yeah that sounds good. I'll leave this PR hanging here for now, until further discussion happens.

PR #38, opened 2020-07-17
Clean up API URLs

mvadari:
Use nested paths to make the `urls.py` file neater and with less repeated text. 

Test:
* derped around the frontend and made sure the modified APIs still worked.


mvadari:
> As an aside, I'm noticing that when I return to the student dashboard, it hits <prog>/student/schedule before <prog>/student. This seems like a bug?

That appears to be a factor of Javascript's asynchronous nature. In the code, `/student` technically is sent first.

PR #40, opened 2020-07-17
Nest React routes

mvadari:
Refactored all the program-specific React routes so that they are nested

Test:
* Went to all the affected routes and made sure they worked properly

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -45,6 +45,10 @@ const NotFound = () => (
   </section>
 );
 
+const Empty = ({ children }) => {
+  return children;
+};
+
 function App(props: {}) {
   const [userInfo, setUserInfo] = useState({
     username: "",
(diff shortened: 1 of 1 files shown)


client/src/App.tsx
```diff
@@ -90,27 +94,32 @@ function App(props: {}) {
...
+              url={liabilityWaiverEndpoint}
+              formName="Liability Waiver Form"
+            />
+            <DummyForm
+              path="availability"
+              url={studentAvailabilityEndpoint}
+              formName="Program Availability"
+            />
+            <ChangeClassesDashboard path="changeclasses" />
+            <NotFound default />
```
wishdasher, line 121: This shouldn't be part of the nesting.

mvadari: I put it in because it wasn't showing up when I went to a URL under the nesting that wasn't explicitly covered (it would just be a blank page).

wishdasher: But the original one should have been left alone.

wishdasher:
Also, in general, these subroutes should be moved out of `App.tsx`.

client/src/App.tsx
mvadari: Is the original one not there? It's supposed to be.

wishdasher: I wasn't looking for it in the right place, sorry, ignore.

mvadari:
> Could you hold off on this? My PR with refactoring student registration takes care of this.

Sure.

mvadari:
Deleting because PR #42 handles this.

PR #41, opened 2020-07-17
Create some initial tests

mvadari:
Yay testing!

Set up some initial tests, with a framework for how to continue writing tests. Probably need some more framework for some sort of integration testing as well, but I think API testing takes care of most of that. 

Thoughts on name formatting for test cases?

Test:
* uhhh ran the tests


wishdasher:
Tests should go into a separate directory (maybe as a sibling of `core`?). It would also be great if there were a way to run all tests at once that could be invoked by a single command.

mvadari:
> It would also be great if there were a way to run all tests at once that could be invoked by a single command.

There is - `./manage.py tests`. It covers all files that start with `test_`.

evantey14:
> Tests should go into a separate directory (maybe as a sibling of `core`?). It would also be great if there were a way to run all tests at once that could be invoked by a single command.

I think the most proper django way would be to make a Python module `tests` that's a child of core (since they're the tests associated with the app core). But yeah either of these is probably preferred over a hanging `test_models` file.

evantey14:
> > It would also be great if there were a way to run all tests at once that could be invoked by a single command.
> 
> There is - `./manage.py tests`. It covers all files that start with `test_`.

I think this should be `test` not `tests`. Anyhow, this is something that should be added to the PR workflow if anyone wants to drop a trello card

mvadari:
> Tests should go into a separate directory (maybe as a sibling of core?).

Done. 

> Anyhow, this is something that should be added to the PR workflow if anyone wants to drop a trello card

Done.

mvadari:
> test_url seems like a good thing we'd want to test, but also as far as I can tell, that code (url) isn't actually being used?

`url` is used in the dashboard API call.

mvadari:
> So, if we can, I think I'd like to start off tests on as good a footing as possible for future tests.

The idea of this PR was to literally just set up the tests pipeline, with some sample tests. I was going to start writing actual tests in following PRs. I added a bunch of placeholders, so hopefully that makes it more clear what I anticipate this looking like (basically mirroring the `models/` and `views/` folders).

server/core/tests/models/program.py
```diff
@@ -0,0 +1,42 @@
...
+
+        # open student registration for one object
+        program = Program.objects.get(name="Droplet", edition="3030")
+        program.student_reg_open = True
+        program.save()
+
+        # make one program a past program
+        program = Program.objects.get(name="Slosh", edition="420")
+        program.student_reg_status = RegStatusOptions.POST_PROGRAM
+        program.save()
```
evantey14, line 18: Can we just pass these in when the programs are defined like `Program.objects.create(name="Droplet", edition="3030", student_reg_open=True)`?

server/core/tests/models/program.py
```diff
@@ -0,0 +1,42 @@
...
+        program = Program.objects.get(name="Droplet", edition="3030")
+        self.assertEquals(program.url, "Droplet/3030")
+
+    def test_open_student_programs(self):
+        open_student_programs = Program.get_open_student_programs()
+        self.assertEquals(len(open_student_programs), 1)
+
+        open_student_program = open_student_programs[0]
+        expected_program = Program.objects.get(name="Droplet", edition="3030")
+        self.assertEquals(open_student_program, expected_program)
```
evantey14, line 30: Are we able to just check that open_student_programs is `[expected_program]` instead of having the two checks?

server/core/tests/models/program.py
mvadari: Oops forgot I could do that.

server/core/tests/models/program.py
mvadari: If you convert the QuerySet to a list, yes.

mvadari:
> I think this structure looks good, though I'd prefer we not just add in a bunch of empty shell structures until we have tests for them (particularly in case we decide to use something else for testing).

I'd like to leave the shell structure in place to make it easier to add tests, but it's also v easy to remove if we decide to go with something else.

PR #43, opened 2020-07-18
Extract login and logout logic from UI components

wishdasher:
Don't mix API calls and localstorage / header mutations with application state changes.

Tested logging in and out.

Taking suggestions for better filename than `manage.ts`.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -80,6 +80,7 @@ function App(props: {}) {
       <main className="px-3 py-5">
         <Router>
           <Home path="/" />
+          {/* TODO: better page than homepage? Notification of logout success? */}
           <Home path="logout" />
           <LoginPage path="login" username={userInfo.username} setToken={setToken} />
           <SignupPage path="signup" username={userInfo.username} setToken={setToken} />
--- a/client/src/layout/Nav.tsx
+++ b/client/src/layout/Nav.tsx
@@ -4,7 +4,7 @@ import "./Nav.css";
 import { Link } from "@reach/router";
 
 import { canonicalizeProgramName, programList } from "../info/Program";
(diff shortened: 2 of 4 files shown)


client/src/layout/Nav.tsx
```diff
@@ -22,12 +22,8 @@ function Nav(props: Props) {
...
   }
 
-  function logout() {
-    console.log("Logging out");
-    localStorage.removeItem("token");
-    localStorage.removeItem("refresh");
-    delete axiosInstance.defaults.headers.common["Authorization"];
-
+  function handleLogout() {
+    logout();
```
evantey14, line 26: Can we leave a TODO or something saying there's probably error handling logic here? (or just idk explain why we have `handleLogout` and aren't just putting `logout` directly in the Link below.

client/src/App.tsx
```diff
@@ -80,6 +80,7 @@ function App(props: {}) {
       <main className="px-3 py-5">
         <Router>
           <Home path="/" />
+          {/* TODO: better page than homepage? Notification of logout success? */}
           <Home path="logout" />
```
mvadari, line 84: Instead of having a page for logout, could there just be a redirect back to the home page?

client/src/layout/Nav.tsx
mvadari: ++

wishdasher: There's the following `setToken` line?

wishdasher: Added a `NOTE`

client/src/App.tsx
wishdasher: That's what it currently does.

PR #45, opened 2020-07-18
Refactor settings into modules

evantey14:
- Move settings into base, local and prod (default to using local)
- Clean out the cookie-cutter django comments
- Add some comments explaining some settings decisions
- Left TODOs for things that should be cleaned up in the future.

Testing:
- Made sure things ran locally and in heroku
- Confirmed the correct settings files were being used (locally, `DEBUG=True` means we see debug screens eg when you hit a bad url. in prod, the react template is only loaded in the prod settings, so it won't work otherwise)


server/config/settings/production.py
```diff
@@ -0,0 +1,31 @@
...
+PROJECT_ROOT = os.path.dirname(BASE_DIR)
+
+DEBUG = False
+SECRET_KEY = os.environ["SECRET_KEY"]
+
+TEMPLATES[0]["DIRS"].append(os.path.join(PROJECT_ROOT, "client", "build"))
+
+# TODO: this probably should pull from env vars
+DATABASES["default"].update(
+    {"USER": "postgres", "PASSWORD": "postgres",}
```
mvadari, line 19: Can we put the username and password in heroku's settings instead of this file for prod? Just to set up that pipeline, even if it is still `postgres`/`postgres`

server/config/settings/production.py
```diff
@@ -0,0 +1,31 @@
...
+)
+
+AUTH_PASSWORD_VALIDATORS = [
+    {"NAME": "django.contrib.auth.password_validation.UserAttributeSimilarityValidator",},
+    {"NAME": "django.contrib.auth.password_validation.MinimumLengthValidator",},
+    {"NAME": "django.contrib.auth.password_validation.CommonPasswordValidator",},
+    {"NAME": "django.contrib.auth.password_validation.NumericPasswordValidator",},
+]
```
mvadari, line 27: Is there a reason these aren't in `local`?

server/manage.py
```diff
@@ -5,7 +5,7 @@
 
 
 def main():
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "config.settings")
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "config.settings.local")
```
mvadari, line 8: Shouldn't this draw from prod for heroku? Or does heroku take care of that?

server/setup_sample_data.py
```diff
@@ -2,7 +2,7 @@
 
 import os  # isort:skip # NOQA
 
-os.environ.setdefault("DJANGO_SETTINGS_MODULE", "config.settings")  # isort:skip # NOQA
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "config.settings.local")  # isort:skip # NOQA
```
mvadari, line 5: Shouldn't this draw from prod for heroku? Or does heroku take care of that?

server/config/settings/production.py
evantey14: Yeah so I was trying to do this but couldn't figure it out, so I left the TODO on line 17 since I didn't want to spend more time on it. Definitely should be done before going public if that ever happens

server/config/settings/production.py
evantey14: I figured these were more annoyances than anything else if you're doing local dev. If you think we should have this, lmk and I can move it back.

server/manage.py
evantey14: Yeah this line says "set `DJANGO_SETTINGS_MODULE` equal to the relevant environment variable if it exists and if it doesn't, set it to `config.settings.local`". In prod, this environment variable is set so it does pull the right settings.

server/setup_sample_data.py
evantey14: (see other comment)

PR #46, opened 2020-07-21
Add PrivateRoute component and redirect correctly

wishdasher:
Routes wrapped behind `PrivateRoute` will redirect to the login if user is not logged in. Upon logging in, the user is redirected to the original page they were viewing.
Currently, only `/nextup` is protected, for testing purposes.

Testing:
- not logged in, access `/nextup`, log in and verify that you are on the correct page
- access `/nextup` while logged in, verify that you are directed straight there
- log in normally, verify that you are redirected to the dashboard (default)

### DIFF
--- a/client/src/PrivateRoute.tsx
+++ b/client/src/PrivateRoute.tsx
@@ -1,0 +1,17 @@
+import { navigate } from "@reach/router";
+import React, { useEffect } from "react";
+
+import { useLoggedIn } from "./context/auth";
+
+//@ts-ignore
+function PrivateRoute({ as: Component, ...props }) {
+  const loggedIn = useLoggedIn();
+  useEffect(() => {
+    if (!loggedIn) {
+      navigate("/login", { state: { referer: props.location } });
+    }
+  }, [loggedIn, props.location]);
+  return <Component {...props} />;
+}
+
+export default PrivateRoute;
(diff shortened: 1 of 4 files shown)


client/src/PrivateRoute.tsx
```diff
@@ -0,0 +1,17 @@
+import { navigate } from "@reach/router";
+import React, { useEffect } from "react";
+
+import { useLoggedIn } from "./context/auth";
+
+//@ts-ignore
```
evantey14, line 6: Is the goal to eventually remove this? (if so, pls add comment about how it's temporary)

client/src/PrivateRoute.tsx
```diff
@@ -0,0 +1,17 @@
...
+function PrivateRoute({ as: Component, ...props }) {
+  const loggedIn = useLoggedIn();
+  useEffect(() => {
+    if (!loggedIn) {
+      navigate("/login", { state: { referer: props.location } });
+    }
+  }, [loggedIn, props.location]);
```
evantey14, line 13: Why does this change on props.location change?

wishdasher: Automatic dependency added by eslint. In theory not relevant.

wishdasher:
Changed `referer` to `referPath`

`props.location`: https://reactrouter.com/web/api/location
`.state`: https://reach.tech/router/api/navigate

evantey14:
> Changed `referer` to `referPath`
> 
> `props.location`: https://reactrouter.com/web/api/location
> `.state`: https://reach.tech/router/api/navigate

Ok, yeah I saw these, but I think I was thrown off by the different uses of `referer` (/ it just took me some time to figure out what was going on).

PR #47, opened 2020-07-21
Set page titles on all accessible pages

wishdasher:
In this SPA of ours, the page title doesn't change when the user navigates. Bad for accessibility, bad for navigation purposes.

This current solution is pretty bad because we have to manually mark every component that could directly routed to (all the "pages"). An alternate solution is to create a HOC `withTitle` that wraps all the components wherever we have routing (like in `App.tsx`). However, this is less flexible and means we won't be able to have dynamic page titles like "Splash 2019 Registration." Instead, it would just be "Program Registration" or something.

I could potentially go with a hybrid solution (`withTitle` HOC on the static pages, which will probably be server-rendered at some point in the future anyways, and custom title setting for the program specific pages).

### DIFF
--- a/client/src/accounts/LoginPage.tsx
+++ b/client/src/accounts/LoginPage.tsx
@@ -10,6 +10,8 @@ type Props = {
 };
 
 function LoginPage(props: Props) {
+  document.title = "Login | MIT ESP";
+
   const loggedIn = useLoggedIn();
 
   return (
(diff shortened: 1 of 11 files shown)


wishdasher:
#49 overrides this PR and uses a different solution. Keeping this PR for context.

mvadari:
You should probably close the PR.

PR #48, opened 2020-07-21
Refactor signup page and form to use new schema

wishdasher:
Testing:
- navigating to signup page while logged in should display a prompt to log out
- make sure that signup process works and is successful and redirects to the dashboard

### DIFF
--- a/client/src/accounts/SignupPage.tsx
+++ b/client/src/accounts/SignupPage.tsx
@@ -1,68 +1,87 @@
-import React, { Component } from "react";
+import { Link } from "@reach/router";
+import React, { useState } from "react";
 
 import StudentSignupForm from "./StudentSignupForm";
 import { SignupType } from "./types";
 
+import { logout } from "../accounts/manage";
+import { useAuth, useLoggedIn } from "../context/auth";
+
 type Props = {
-  username: string;
   setToken: (token: string) => void;
 };
 
-type State = {
-  selectedSignupType: SignupType;
-};
+function SignupPage(props: Props) {
+  const { username } = useAuth();
+  const loggedIn = useLoggedIn();
 
-class SignupPage extends Component<Props, State> {
-  constructor(props: Props) {
-    super(props);
-    this.state = {
-      selectedSignupType: SignupType.Student,
-    };
+  const [selectedSignupType, setSelectedSignupType] = useState(SignupType.Student);
+
+  function getButtonClass(signupType: SignupType) {
+    return `button ${selectedSignupType === signupType ? "is-active" : ""}`;
   }
 
-  getButtonClass(signupType: SignupType) {
-    return `button ${this.state.selectedSignupType === signupType ? "is-active" : ""}`;
-  }
-
-  handleClick(signupType: SignupType) {
-    if (signupType !== this.state.selectedSignupType) {
-      this.setState({ selectedSignupType: signupType });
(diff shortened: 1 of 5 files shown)


client/src/accounts/SignupPage.tsx
```diff
@@ -1,68 +1,87 @@
...
+    if (signupType !== selectedSignupType) {
+      setSelectedSignupType(signupType);
     }
   }
 
-  render() {
+  function handleLogout() {
+    logout();
+    // NOTE: we need to set the token to empty for application context state
+    props.setToken("");
```
evantey14, line 33: This seems like a recurring thing? Will there be times we want to `logout()` but _don't_ want to `setToken` to empty? If not, I think it'd be nice to tie these actions together somehow (though seems hard since setToken gets passed down?).

wishdasher: That was the point of `manage.ts`, which separates the server changes from the application state changes. Unless I make another React component just to encapsulate these two lines, they have to be called together.

wishdasher:
> One thing to consider that might things more clean is wrapping all the profile info into a single object? Seems like this may cut down on the long field lists that this points towards (e.g in the switch case and in the signupStudent function). Don't think it's too critical now, but we definitely should think about this before just adding more fields to the form.

I don't think a single object avoids long field lists. TS would still force you to list out all the existing fields. So that is not the solution.

> Maybe unrelatedly, I'm getting an error when I create a student? After clicking "Submit" I get redirected to an empty dashboard (no dashboard components seem to load?). When I refresh, everything then loads and looks correct. This may have been an existing problem? It appears to me like App's useEffect isn't being triggered? Could this be fixed by changing out handleSignup works? (Should this be a new PR?)

Known bug: https://trello.com/c/8HRoAVI0/89-bug-rendering-sometimes-fails-when-navigating-by

client/src/accounts/SignupPage.tsx
evantey14: Hmmm ok yeah that separation does make sense. We'll have to watch out for this, though, cause unsuspecting people (like me probably) may forget / not realize `logout` doesn't handle everything related to logging out.

evantey14:
> > One thing to consider that might things more clean is wrapping all the profile info into a single object? Seems like this may cut down on the long field lists that this points towards (e.g in the switch case and in the signupStudent function). Don't think it's too critical now, but we definitely should think about this before just adding more fields to the form.
> 
> I don't think a single object avoids long field lists. TS would still force you to list out all the existing fields. So that is not the solution.

Er yeah, so it wouldn't remove long lists permanently, but as it is now, if you wanted to add a field, you'd have to make changes at 5-6 different places whereas I think if wrapped into some sort of profile object, you'd probably be able to reducing this to maybe 2 locations? Anyhow, definitely not something we need to change now.

> > Maybe unrelatedly, I'm getting an error when I create a student? After clicking "Submit" I get redirected to an empty dashboard (no dashboard components seem to load?). When I refresh, everything then loads and looks correct. This may have been an existing problem? It appears to me like App's useEffect isn't being triggered? Could this be fixed by changing out handleSignup works? (Should this be a new PR?)
> 
> Known bug: https://trello.com/c/8HRoAVI0/89-bug-rendering-sometimes-fails-when-navigating-by
Ah -- great (Y)

PR #49, opened 2020-07-21
Add Page HOC and apply to all reachable pages

wishdasher:
This adds a higher-order page component. `Page.tsx` exports three functions meant to wrap page-level components (components that have routes).

There are a few reasons to do this:
- standardize how pages are constructed, which makes things like spacing consistent. Useful in the future if we need to add site-wide banner components, make the shared layout more complex, or something similar.
- in some cases, decrease nesting in render functions
- make the changes from #47 feel less scattered. Now, the setting of `document.title` always happen in the page functions.
- allow for future extensiblity of the `Page` component. Shortlink tracking, whatever.

Weakness:
- still need to remember to use a page function for any "page" component. Optimally, we would be able to do this wrapping at the router level, but Reach Router is strict about what its children could be, and I wasn't clever enough to come with a different solution.

Testing:
- went to all the pages and made sure they looked the same. In some cases, they were improved because the original version had left out the `container` class on the outermost div.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -19,16 +19,18 @@ import Teach from "./info/Teach";
 import Footer from "./layout/Footer";
 import Nav from "./layout/Nav";
 import PrivateRoute from "./PrivateRoute";
+import { contentPage } from "./layout/Page";
 import RegDashboard from "./registration/RegDashboard";
 
-const NotFound = () => (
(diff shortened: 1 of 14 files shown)


wishdasher:
In terms of reviewing, only `Page.tsx` is interesting, and maybe look at a couple of the modified components to see how the wrapper functions were applied.

PR #51, opened 2020-07-22
Only retrieve schedule if a valid regstatus

mvadari:
The schedule is currently always retrieved in the reg dashboard. This is an edit to ensure that it is only retrieved if the student has a valid registration status.

Test:
* checked whether the API call was made based on the reg status
* made sure the schedule rendered correctly if in a valid reg status


evantey14:
I'm gonna say we punt this until tonight when we can talk in more detail about student registration flow

PR #52, opened 2020-07-22
Resolve UpdateProfileForm errors

mvadari:
In response to [this Trello card](https://trello.com/c/zo7BZdRA/91-bug-updateprofileform-violates-various-reacty-things)

Test:
* Opened the "updateprofile" page and made sure the warnings no longer appeared
* Made sure the form still submitted the GET/POST correctly


wishdasher:
I don't think this is the correct approach, or at least, it should be discussed first and applied more comprehensively.

Does this imply that every field for every model should always have a default value?

mvadari:
> I don't think this is the correct approach, or at least, it should be discussed first and applied more comprehensively.
> 
> Does this imply that every field for every model should always have a default value?
> 
> @evantey14

I think the comment above `Profile` addresses this. A lot of the defaults/blanks in the user-related models are because we want to be lazy on account creation and stuff, as we test everything. The TODO there is a note that at some point those all need to be revisited, and we need to actually think through everything there/what would be a good default value. 

The bug won't occur if the dob and grad year are correctly formatted; it's only if it's whatever the empty value is (which is null). Since this is a temporary measure, I picked default values that are not good real-world examples.

evantey14:
Hmm is the purpose of this just to hotfix the bad form so we can better test student registration flow? If so, then I guess this is fine for now, but @wishdasher yeah lots of things about fields should get cleaned up. e.g. The fields shouldn't have default values. I also think they should be renamed (e.g. grad_year -> graduation_year). There's a lot of validation we should have. The textfield lengths should also be chosen more intentionally.

For the record, I'm still pretty strongly of the opinion that we should just get rid of most of these fields for now. Adding new fields once we've figured out how to handle them cleanly is much easier than changing lots of code. I think leaving placeholders like an isEligible function that just returns true (or does a naive hard-coded check like "are you in high school") is fine / the marginal gain of that functionality comes at the cost of quite a bit of code that could quickly get stale (in the model, in front end forms, etc)

wishdasher:
Oh yeah, can we just get rid of these fields for now? Including the two on the student sign up form.

mvadari:
Updated @wishdasher  @evantey14

evantey14 (commented):
I feel like this is missing some changes? It looks like all that got removed is the rendering and existence in the database, but what about wherever the axios request is being sent? 

@wishdasher are you asking that we get rid of all the fields on the update profile form and signup forms? (cause I'd be heavily in favor of that and replacing update profile with either a nonfunctional form of just dummy submit button). If yes, we should probably put it into a new PR.

mvadari:
Oops - everything has been commented out now.

PR #53, opened 2020-07-23
Fix class changes dashboard

mvadari:
* Fixed links so that everything routed correctly
* Converted class changes to a function from a class (this happened before I realized the former was the issue, and I figured we'd need it to happen later anyways)

Test:
* Went to /changeclasses and made sure it rendered correctly/did everything it was supposed to
* Tested all the changed links and made sure they routed correctly

### DIFF
--- a/client/src/registration/ChangeClassesDashboard.tsx
+++ b/client/src/registration/ChangeClassesDashboard.tsx
@@ -1,4 +1,4 @@
-import React, { Component } from "react";
+import React, { useCallback, useEffect, useState } from "react";
 
 import { Class, ScheduleItem, Section } from "./types";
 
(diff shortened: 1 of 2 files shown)


client/src/registration/ChangeClassesDashboard.tsx
```diff
@@ -26,49 +26,38 @@ type State = {
...
-      .then(res => {
-        this.setState({
-          catalog: res.data,
-          displayedCatalog: res.data,
-        });
-      });
-  }
+  useEffect(() => {
+    // set up student schedule
+    setupStudentClasses();
```
wishdasher, line 52: Why separate this out?

mvadari: It's used when classes are added/removed, to refresh the schedule post-change.

PR #54, opened 2020-07-23
Reorganize frontend program-specific links

mvadari:
* Move program-specific links to a separate file
* Change exporting method in `apiEndpoints` (so there didn't need to be another list of variables at the bottom)

Test:
* Went to all the pages, made sure they rendered correctly
* Checked all the links

### DIFF
--- a/client/src/apiEndpoints.ts
+++ b/client/src/apiEndpoints.ts
@@ -1,44 +1,26 @@
 // Accounts
-const loginEndpoint = "/token/";
-const tokenRefreshEndpoint = "/token/refresh/";
-const userDataEndpoint = "/user/";
-const studentSignupEndpoint = "/account/student/";
+export const loginEndpoint = "/token/";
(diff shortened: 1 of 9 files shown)


wishdasher:
> Move program-specific links to a separate file
> Rename dashboards so they are ProgramDashboard instead of RegDashboard
> Rename the folder to program, which is more descriptive

I did the last two as well in my current changes. For the sake of avoiding really painful rebases, would you be okay not merging this? Or only merging the first point. Sorry for not being more communicative of what I've been doing.

mvadari:
> > Move program-specific links to a separate file
> > Rename dashboards so they are ProgramDashboard instead of RegDashboard
> > Rename the folder to program, which is more descriptive
> 
> I did the last two as well in my current changes. For the sake of avoiding really painful rebases, would you be okay not merging this? Or only merging the first point. Sorry for not being more communicative of what I've been doing.

Done. I added another minor change instead that I've been meaning to do, because it's in line with the objective of this PR.

PR #55, opened 2020-07-23
Make program-specific routes API-based

mvadari:
Currently, the website allows you to go to any URL that is of the form `/[something]/[something2]` and it will treat it as a program. This means that people can try to go to `Splash 2040` or `Slosh 420` or `sijdfoiasd asoidfjaois` and our website will render it as a program. 

This PR pulls programs from the backend, to ensure that the only viable routes are the ones that are actually programs. 

Test:
* Went to all the different pages for an actual program, made sure they rendered correctly (with their API calls)
* Tried going to the page of a nonexistent program, verified that it showed up as `404 Not Found`

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -59,7 +59,7 @@ function App(props: {}) {
     axiosInstance.get(programsEndpoint).then(result => {
       setPrograms(result.data);
     });
-  });
+  }, []);
 
   function setToken(accessToken: string) {
     setAuthToken(accessToken);
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -5,7 +5,7 @@ import React, { useEffect, useState } from "react";
 
 import LoginPage from "./accounts/LoginPage";
 import SignupPage from "./accounts/SignupPage";
-import { userDataEndpoint } from "./apiEndpoints";
+import { userDataEndpoint, programsEndpoint } from "./apiEndpoints";
 import axiosInstance from "./axiosAPI";
 import { AuthContext } from "./context/auth";
 import Dashboard from "./dashboard/Dashboard";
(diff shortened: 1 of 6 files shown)


client/src/App.tsx
```diff
@@ -51,6 +53,9 @@ function App(props: {}) {
         });
       });
     }
+    axiosInstance.get(programsEndpoint).then(result => {
+      setPrograms(result.data);
+    });
```
wishdasher, line 58: This isn't dependent on `authToken`, let's make it a separate `useEffect` call, with second parameter `[]`.

client/src/App.tsx
```diff
@@ -5,21 +5,22 @@ import React, { useEffect, useState } from "react";
...
+import { userDataEndpoint, programsEndpoint } from "./apiEndpoints";
 import axiosInstance from "./axiosAPI";
 import { AuthContext } from "./context/auth";
 import Dashboard from "./dashboard/Dashboard";
 import AboutUs from "./info/AboutUs";
 import Home from "./info/Home";
 import Learn from "./info/Learn";
 import Nextup from "./info/Nextup";
-import Program, { programList } from "./info/Program";
+import ProgramStatic, { programList } from "./info/Program";
```
wishdasher, line 16: No reason to rename this, and I think it's confusing that it's being renamed `ProgramStatic` when being imported from `Program.tsx`.

Just call the other thing `ProgramInfoType` or something.

client/src/registration/RegDashboard.tsx
```diff
@@ -65,9 +66,18 @@ function RegDashboard(props: Props) {
               edition={props.edition}
             />
             {/* @ts-ignore TODO: reach-router path fix */}
```
wishdasher, line 68: You can move this line down one. It's `path` that's the issue. Same for other change below.

client/src/App.tsx
mvadari: Ah ok, I thought you could only have one `useEffect` call

client/src/registration/RegDashboard.tsx
mvadari: Ah that's what that's for. Done.

client/src/App.tsx
evantey14: lmao same^ until 2 days ago when I was working on something else and spent 3 hours trying to be clever about it only to realize I can just have different useEffects :(

wishdasher:
This PR breaks authentication. When you open the app without being logged in, you get infinite 401s from the requests to `http://localhost:8000/api/user/ ` and `http://localhost:8000/api/programs/`.

Make sure the homepage (nav bar with sign up / log in buttons) show up correctly when testing the fix.

PR #60, opened 2020-07-24
Make program-specific routes API-based

mvadari:
Try number 2 for PR #55 

Problem fixed (I believe)

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -59,7 +59,7 @@ function App(props: {}) {
     axiosInstance.get(programsEndpoint).then(result => {
       setPrograms(result.data);
     });
-  });
+  }, []);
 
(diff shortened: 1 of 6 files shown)


client/src/App.tsx
```diff
@@ -53,6 +55,12 @@ function App(props: {}) {
     }
   }, [authToken, props]);
 
+  useEffect(() => {
+    axiosInstance.get(programsEndpoint).then(result => {
```
wishdasher, line 59: Is there a reason to use `axiosInstance` at all, instead of a bare `axios`?

mvadari: The beginning of the URL (which is dependent on dev vs prod) is stored in the instance.

wishdasher (approved):

PR #61, opened 2020-07-27
Start new registration and program information display flow

wishdasher:
New stuff at `beta/:program/:edition`. Copied over necessary parts from certain files in `client/src/registration`.

Mostly restructured how the components worked with each other, separated out teacher and student program pages.

Next:
- actually enforce a registration steps -> program page info flow
- add in general program information to the relevant pages
- protect certain URLs and redirect logged in / not logged in traffic accordingly
- delete old files from `registration`

### DIFF
--- a/client/src/program/GeneralStudentPage.tsx
+++ b/client/src/program/GeneralStudentPage.tsx
@@ -1,0 +1,140 @@
+import { Link } from "@reach/router";
+import React, { useEffect, useState } from "react";
+
+import { studentScheduleEndpoint } from "../apiEndpoints";
+import axiosInstance from "../axiosAPI";
+import { renderLinkedText, renderTextInSection } from "../helperTextFunctions";
+import { generalPage } from "../layout/Page";
+import { RegStatusOption, ScheduleItem } from "../registration/types";
+
+type Props = {
+  program: string;
+  edition: string;
+  checks: {
+    availabilityCheck: boolean;
+    emergencyInfoCheck: boolean;
+    liabilityCheck: boolean;
+    medliabCheck: boolean;
+    updateProfileCheck: boolean;
+  };
+  regStatus: RegStatusOption;
+};
+
+function GeneralStudentPage(props: Props) {
+  const [schedule, setSchedule] = useState([] as ScheduleItem[]);
+
+  function renderRegistrationLink() {
+    return <Link to="register">Registration steps</Link>;
+  }
+
+  useEffect(() => {
+    // Set up student classes
+    axiosInstance.get(`/${props.program}/${props.edition}/${studentScheduleEndpoint}`).then(res => {
+      setSchedule(res.data);
+    });
+  }, [props.edition, props.program]);
+
+  function renderClassPrefs(editsAllowed: boolean = false) {
+    return (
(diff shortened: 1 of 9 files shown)


client/src/program/ProgramPage.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const { program, edition } = props;
+
+  return (
+    <React.Fragment>
+      {isStudent ? (
+        <StudentProgramPage program={program} edition={edition} />
+      ) : isTeacher ? (
+        <TeacherProgramPage program={program} edition={edition} />
+      ) : (
+        <StudentProgramPage program={program} edition={edition} />
```
mvadari, line 24: What's going on here? Why is the `StudentProgramPage` being displayed if the user is neither a student nor a teacher? If this is a placeholder, can you add a `TODO`?

client/src/program/GeneralStudentPage.tsx
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+
+  function renderClassStatus() {
+    const canAddClasses =
+      props.checks.updateProfileCheck &&
+      props.checks.emergencyInfoCheck &&
+      props.checks.medliabCheck &&
+      props.checks.liabilityCheck &&
+      props.checks.availabilityCheck;
```
mvadari: Should this be replaced with `isRegistered` or will that include things like payment checks?

client/src/program/StudentProgramPage.tsx
```diff
@@ -0,0 +1,69 @@
...
+        });
+        setRegStatus(res.data.reg_status);
+      });
+    }
+  }, [loggedIn, props.edition, props.program]);
+
+  return (
+    <React.Fragment>
+      <Router primary={false}>
+        <GeneralStudentPage
```
mvadari: This naming scheme is confusing. Until I saw this line, I was confused about what the difference was between `StudentProgramPage` and `GeneralStudentPage`.

client/src/program/helpers.ts
```diff
@@ -0,0 +1,6 @@
+function isStudentRegistered(regChecks: object) {
+  //@ts-ignore
+  return Object.keys(regChecks).every(k => regChecks[k]);
+}
+
+export { isStudentRegistered };
```
mvadari: If this isn't being used right now, I would take it out and put it back in when it is being used.

mvadari (changes requested):

client/src/program/ProgramPage.tsx
wishdasher: We talked about student view potentially being a default, but `TODO` for now.

client/src/program/GeneralStudentPage.tsx
wishdasher: Nope. Because we can be in an intermediate state of registration, will be useful later.

client/src/program/StudentProgramPage.tsx
wishdasher: Renamed to `StudentProgramPageRouter` (because that's its functionality) and `StudentProgramInfoPage`

